PR #150, opened 2013-02-28
Added luhn validator.

xavierzwirtz:
Added a luhn validator to parsley.extend.js. Uses luhn algorithim to
verify that fields value is correct. Would be used on credit card fields
to assert that they have a valid credit card number, but can be used on
other types of data as well.

### DIFF
--- a/documentation.html
+++ b/documentation.html
@@ -1341,6 +1341,13 @@ data-message="You must enter a 10 characters alphanumeric value"</code></pre>
                           Validates that a field's value is present within the value list. You can define the delimiter using <code>data-inlist-delimiter=","</code>. Delimiter defaults to <code>","</code>.
                         </td>
                       </tr>
+                      <tr>
+                        <td>Luhn</td>
+                        <td><code>data-luhn="true"</code></td>
+                        <td>
+                          Validates that a fields value passes the Luhn algorithm. Validates credit card numbers, as well as some other kinds of account numbers.
+                        </td>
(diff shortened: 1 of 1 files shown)


custa1200:
How could this be extended to not only include a valid credit card string but to also somehow add the card type name to style inputs with card images.

marcandre:
I'm not sure this is related to *validation* per say, plus it should be quite easy to do with custom JS code.

custa1200:
A credit card input could be seen as meeting the Luhn algorithm yet it does not mean it is actually a valid Credit card, i.e. starting with the number 5 and meeting the Luhn does not mean that it is a valid Mastercard.

marcandre:
Personally, I dislike when sites ask me the type of credit card I have, as it's entirely redundant. In any case, you're welcome to do that, but there's no easy way to add that validation.
- good websites don't ask for CC type
- those that do won't have a standard way to input the CC type
- moreover, parsley doesn't currently deal well with dependencies.

voiceofwisdom:
@custa1200  This sounds out of scope for parsley. [This seems to be doing exactly what you want.](http://jquerycreditcardvalidator.com/)

guillaumepotier:
I totally agree with @voiceofwisdom, seems that kind of 'validation' is way too specific and should be implemented by end user, relying if they want on `luhn` validator. Jquery plugin mentioned above seems to do the job too.

PR #349, opened 2013-08-28
Added scrollDuration option...

nicooprat:
...to scroll smoothly to the first focused invalid field instead of jumping at it. Focus is set at the end of the scroll.

Hope it helps.

### DIFF
--- a/parsley.js
+++ b/parsley.js
@@ -1194,7 +1194,11 @@
         }
       }
 
-      this.options.listeners.onFormSubmit( valid, event, this );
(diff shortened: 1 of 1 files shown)


eladvider-softangible:
this is not working now

nicooprat:
Wow, this one is old! Looks like the plugin has been entirely rewritten. You should take a look here:

https://github.com/guillaumepotier/Parsley.js/blob/master/src/parsley/ui.js#L63

And try to copy/paste/adjust my old code:

https://github.com/guillaumepotier/Parsley.js/pull/349/files

Issue #644, opened 2014-05-15
Proposals for depending on other fields validators


Jako:
Parsley misses validators for fields depending on each other. This is just a proposal how that could be done – the parameters are not universal enough in my opinion.

```
    // example: data-parsley-requiredonradio="usertype,company"
    window.ParsleyValidator
        .addValidator('requiredonradio', function (value, requirements) {
            requirements = requirements.split(',');
            // if the value of the radio with name 'requirements[0]' is equal to requirements[1], field is required
            if (requirements[1] == $('[name="' + requirements[0] + '"]:checked').val() && '' == value) {
                return false;
            }
            return true;
        }, 32)

    // example: data-parsley-requiredoncheckbox="delivery"
    window.ParsleyValidator
        .addValidator('requiredoncheckbox', function (value, requirement) {
            // if checkbox with name 'requirement' is checked, field is required
            if ($('[name="' + requirement + '"]').prop('checked') && '' == value) {
                return false;
            }
            return true;
        }, 32)
```

marcandre:
Related to #839

pljspahn:
An example based on something I am currently working on (and deciding if Parsley will meet the needs of) is validating multiple `select` elements that allow a user to input a dimension in inches with fractional values.

	<select name="inch" id="inch">
		<option value="1">1</option>
		<option value="2">2</option>
		...
		<option value="50">50</option>
	</select>
	<select name="fraction" id="fraction">
		<option value="0">0</option>
		<option value="0.125">1/8</option>
		<option value="0.25">1/4</option>
		...
		<option value="0.875">7/8</option>
	</select>

This form has a handful of inputs like this. One for outer width, outer height, inner width, inner height, etc ... Validation involves, for example, the inner width being less than the outer width (and vice versa). 

So ultimately, the two values from multiple `select` elements need to be combined and compared to the combined values of two other `select` elements.

PR #714, opened 2014-08-19
added extra/depending.js validator (Refs #644)

Jako:


marcandre:
Thanks for the PR.

I'm not sure this feature is the right approach though.

I would simply bind on the radio button and change the `required` attribute accordingly. Moreover, since this would potentially clear/set a validation error, some extra work is probably required (setting the `trigger` option or triggering the right event).

This feature seems too specialized to be useful in the general scheme of things.

PR #731, opened 2014-09-13
Use number constraint for HTML5 number inputs if an appropriate 'step' attribute is set

maniacyak:
This change adds a condition that uses the "number" constraint for HTML5 number inputs if the `step` attribute is set to "any" or a non-integer value.

Per the [specification](http://www.w3.org/TR/html5/forms.html#the-step-attribute), floating-point values are allowed in HTML5 number inputs if the [`step` attribute](https://developer.mozilla.org/en/docs/Web/HTML/Element/Input#attr-step) is set to "any" or a non-integer value. 

Parsley incorrectly defaults to using the "integer" constraint for HTML5 number inputs, and this behaviour can't be overridden by setting a `data-parsley-type` attribute on the field.

### DIFF
--- a/src/parsley/field.js
+++ b/src/parsley/field.js
@@ -201,14 +201,17 @@ define('parsley/field', [
       if ('undefined' === typeof type)
         return this;
 
-      // Small special case here for HTML5 number, that is in fact an integer validator
-      if ('number' === type)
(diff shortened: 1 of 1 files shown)


marcandre:
This is a step in the right direction, but leaves some issues. I hope I got them right in #1037

Issue #735, opened 2014-09-19
addAsyncValidator: beforeSend function to hande ajax data option and other who need element state


robgha01:
Fix forked: https://github.com/guillaumepotier/Parsley.js/issues/736

I have the following code:
```addAsyncValidator('validzipcode',
    function (value) {
        console.log(this.$element); // jQuery Object[ input[name="q"] ]

        return 404 === value.status;
    }, "/ServiceMethods.asmx/SsnIsValid", { type: "POST", data: JSON.stringify({ ssn: ??? }), dataType: "json", contentType: "application/json; charset=utf-8" });```

I cant figure out how to pass the value from the input that i have decoreted the validator with to the data -> ssn object i first tried to write "value" and then parsley would change that to the input value but it dident work that way how is it possible to pass the value to ssn without calling $("element").val() ?

To solve this maybe a new attribute can be added where we get the opportunity to set the name of the value that will get sent to the server or to use some sort of magic word that get replaced ex: "parsleyValue" something like that ?

marcandre:
Related to #627

marcandre:
Now fixed, see #894

PR #736, opened 2014-09-21
#735 Added options logic for passing in a function

robgha01:
this allows for dynamicly writing to ajax options with the input value.
issue: https://github.com/guillaumepotier/Parsley.js/issues/735

### DIFF
--- a/src/extra/plugin/remote.js
+++ b/src/extra/plugin/remote.js
@@ -24,15 +24,12 @@ window.ParsleyExtend = $.extend(window.ParsleyExtend, {
     }
   }, window.ParsleyExtend.asyncValidators),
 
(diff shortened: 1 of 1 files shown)


marcandre:
Thanks for the PR.

IIUC, you want to customize the remote options before the AJAX call.

I see that your PR permits this, but if I'm not mistaken, you can already achieve this by listening to the `parlsey:field:validate` event and setting `options.remoteOptions` to your taste, no?

Finally, if we really need another way to do this, it would be best to accept that `remoteOptions` be a function (to be called when needed) instead of adding an extra parameter to `addAsynchValidator`.

In any case, new functionality requires tests.

PR #791, opened 2014-12-12
Field is valid

marcandre:
Return `true` instead of `[]` when there are no validations.
Fix invalid doc for `Field#isValid` and make it clearer when `[]` is returned.


guillaumepotier:
Closed in favor of #808

Issue #805, opened 2014-12-25
Custom message for custom validator


jalandis:
Custom messages not working as expected with custom validators when containing hyphens.  Not sure if this is intended or a bug.
```
<input data-parsley-trigger="submit"
           data-parsley-length="[8, 16]"
           data-parsley-password-complexity="3"
           data-parsley-password-complexity-message="Custom error."
           type="password"
           value="">
```

```
window.ParsleyValidator.addValidator('passwordComplexity', function (value, requirement) {...}, 32);
```

brunowego:
Inside validator, how access data attributes @jalandis? thanks.

brunowego:
Any idea?

marcandre:
@brunowego please realize that it would be best to
1) read the Readme and the Contributing guidelines, and thus ask questions on StackOverflow (as requested)
2) not post in issues that have been closed for months
3) post a question that is understandable, ideally with a working example
This is valid for any open source project, btw...

I also ignore why you ask the question specifically to @jalandis who was simply kind enough to signal a problem and provide a solution.

brunowego:
Ok, thanks @marcandre

PR #806, opened 2014-12-25
Added scenario and removed lowercase of constraint name.

jalandis:
Added support for custom validator messages with hyphens.

Addresses issue #805 

  220 tests complete (954 ms)
  1 test pending

### DIFF
--- a/src/parsley/field.js
+++ b/src/parsley/field.js
@@ -118,7 +118,6 @@ define('parsley/field', [
     * @param {Boolean}  isDomConstraint   optional
     */
     addConstraint: function (name, requirements, priority, isDomConstraint) {
-      name = name.toLowerCase();
 
       if ('function' === typeof window.ParsleyValidator.validators[name]) {
         var constraint = new ConstraintFactory(this, name, requirements, priority, isDomConstraint);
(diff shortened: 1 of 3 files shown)


marcandre:
Looks like a great PR to me.
I doubled checked if there was a valid reason for the `toLowercase` to be there and couldn't find anything. It dates from the first code iteration too.
If I don't hear scream from @guillaumepotier I'll merge this in a couple of days.

guillaumepotier:
I chose originally to lowercase to avoid user typo and ensure that data-Required matched correctly required constraint. Dunno if it is excessive or not. Since it seems more limiting than handy we could try to ship it in 2.1. But it would be BC Break for some users that left case typo.

marcandre:
That's what I thought (avoiding errors).
Sounds good for 2.1 then.

marcandre:
@jalandis If you want to make your PR perfect, you could:
* make the test case title more explicit about compound name
* rebase it off the 2.1 branch.
* modify the CHANGELOG

guillaumepotier:
:+1:

marcandre:
Merged. Thanks so much for the PR, it's much appreciated!

PR #808, opened 2014-12-28
`isValid()` method returns boolean now

guillaumepotier:


guillaumepotier:
I rebased this PR with `2.1` (itself rebased with `master`) to take into account #811. All tests a good. Do you have any feedback/objection before merge, do I still miss something here?

marcandre:
Well, yes, same one I tried explaining earlier. In code:

    // Both results should be the same
    $('<input type="email" value="">').parsley().isValid(false)
    $('<input type="email" value="foo">').parsley().isValid(false, "")

Unless I'm mistaken, the second line returns `false` instead of `true`.

FWIW, I liked having a special return value for the case where validity is unknown (i.e. empty, not required, with constraints), but your proposed API can work too. I think you'll need an explicit `value` argument to `needsValidation` to fix it.

guillaumepotier:
I added some more tests and they pass:

```javascript
      // Value could be passed as argument, needed to add more power to 'parsley:field:validate'
      if ('undefined' === typeof value || null === value)
        value = this.getValue();

      // An empty optional field is valid
      if (!value.length && !this.needsValidation() && true !== force)
        return true;
```

The `!value.length` test in the second statement ensure that a custom value passed as argument is not empty. I'm not sure `needsValidation()` should be aware of this.

More generally, I'm absolutely not fan of this API and that way to pass value in argument through `validate()` to allow value alteration via events. Maybe we should not allow this anymore and instead force users use `data-parsley-value` or directly `value` attribute instead.

marcandre:
> I added some more tests and they pass:
> `expect(true, $('<input type="email" value="foo@bar.baz">').parsley().isValid(false, ""));`

Yeah, that's not a good test though, change the value attribute to `foo` (i.e. an invalid email) and you'll see what I'm talking about!

guillaumepotier:
You were right.. Long time without working on Parsley and javascript more generally, guess I'm a bit rusty right now.. :(

marcandre:
Nice. Apart from a possible simplification, looks good.

PR #809, opened 2014-12-28
Fix field creation problems

marcandre:
TLDNR: This PR takes the approach that Parsley can’t and shouldn't filter fields at the constructor level and changes `init` to always return a `ParsleyForm` or `Parsley{Multiple}Field`, never a naked `Parsley`. It also adds error containers lazily (separate commit, could be split in a different PR).

Details:
In different circumstances, when calling `$.fn.parsley`, I get `Parsley` instances when I want `ParsleyField`.

I do not understand the idea behind return `Parsley` instances; it is definitely not useful for me, but maybe I'm missing something?

In any case, the current algorithm to decide to return a ParsleyField vs Parsley is deeply flawed. Ex:

    <form data-parlsey-excluded=“nothing”><input type=“hidden”></form>
    $(‘form').parsey()
    $(‘input’).parsley() # => Parsley, should be ParsleyField

This is because `Parsley#init` relies on incompletely formed options. Before a `ParsleyField` is created, the `OptionFactory` has no parent set, so won’t inherit the form’s `excluded`.

In any case, I believe that filtering at this stage is not the best approach.

For example, it seems natural to me to be able to add a constraint to a field even if it is currently in the excluded list, to account for the case where it might become included later. E.g:

    <form data-parlsey-excluded=“:hidden”><input type=“text" style=“display: none"></form>
    $(‘form').parsey()
    $(‘input’).parsley().addConstraint(…)
    $(‘.some-button’).click( function() { $(‘input’).show() } )

Currently, the way to achieve this is harder as it requires to subscribe to ‘parsley:form:init’ event and add the constraint at that point. This example is simple, but imagine more complex cases, where the visibility of an input depends on different code, loaded in an order that might change (and could suddenly break things if it hides the field before it’s bound).

In short, trying to decide if an element is a valid input or not in the constructor is tricky, brittle and 
best changed.

There was code to avoid creating a UI for excluded elements. I don’t think the condition was ever false, as parsley:field:init was never emitted for excluded elements.

I kept the code there in the first commit so that excluded fields don’t create UI unless it’s needed. This paves the way to a lazy adding of the error container. This avoids a some DOM pollution, in particular for fields that have no constraints… This last part could be separated in a different PR.

### DIFF
--- a/src/parsley.js
+++ b/src/parsley.js
@@ -72,11 +72,8 @@ define([
       if (this.$element.is('form') || (ParsleyUtils.attr(this.$element, this.options.namespace, 'validate') && !this.$element.is(this.options.inputs)))
         return this.bind('parsleyForm');
 
-      // Every other supported element and not excluded element is binded as a `ParsleyField` or `ParsleyFieldMultiple`
-      else if (this.$element.is(this.options.inputs) && !this.$element.is(this.options.excluded))
-        return this.isMultiple() ? this.handleMultiple(parsleyFormInstance) : this.bind('parsleyField', parsleyFormInstance);
-
-      return this;
+      // Every other element is binded as a `ParsleyField` or `ParsleyFieldMultiple`
+      return this.isMultiple() ? this.handleMultiple(parsleyFormInstance) : this.bind('parsleyField', parsleyFormInstance);
     },
 
     isMultiple: function () {
--- a/src/parsley/ui.js
+++ b/src/parsley/ui.js
@@ -246,12 +246,6 @@ define('parsley/ui', [
       // Store it in fieldInstance for later
       fieldInstance._ui = _ui;
 
-      // Stops excluded inputs from getting errorContainer added
-      if( !fieldInstance.$element.is(fieldInstance.options.excluded) ) {
-        /** Mess with DOM now **/
-        this._insertErrorWrapper(fieldInstance);
-      }
-
       // Bind triggers first time
       this.actualizeTriggers(fieldInstance);
     },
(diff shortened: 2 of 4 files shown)


marcandre:
Rebased onto new 2.1

guillaumepotier:
Okay,

I see more clearly now what bothered me here. In fact, this PR is not _that_ impacting because you allow binding everything with explicit `.parsley()` call.

What I _feared_ was an alteration of `ParsleyForm._bindFields()` that could be disastrous in term of performances if binding everything inside a `<form>` tag ([L96](https://github.com/guillaumepotier/Parsley.js/blob/master/src/parsley/form.js#L96) & [L100](https://github.com/guillaumepotier/Parsley.js/blob/master/src/parsley/form.js#L100)).

But looking into that, the problem of excluded field remains for everyone binding forms and not specifically fields. This PR could be a start but a deeper reflexion may be addressed in the OptionsFactory PR.

Best

marcandre:
Indeed, no change to `_bindFields`.

> But looking into that, the problem of excluded field remains for everyone binding forms and not specifically fields.

Not sure what you mean... Only `<form>` elements can be `ParsleyForm`s, no?

> a deeper reflexion may be addressed in the OptionsFactory PR.

Maybe, but I'm not sure there's anything else to change besides what I'm proposing here.

guillaumepotier:
Not sure what you mean... Only <form> elements can be ParsleyForms, no?

Yes, but when you do `$('form').parsley()`, `ParsleyForm` automatically binds supported inputs and ignore excluded inputs. If you had made your modifications here too Parsley would have bind every html tag inside a form.

My bad from the very beginning. This PR is cool then ;)

marcandre:
Sweet :+1:

PR #810, opened 2014-12-28
Simplifications

marcandre:
Simple simplifications to existing code (and thus no tests)


marcandre:
Cool, so this should be good to merge then, right?

guillaumepotier:
Yes, right too. Sorry again for the delay :s

guillaumepotier:
I'll try to release a new stable version from master during the week.

Best

PR #811, opened 2014-12-29
Fix this.value override to allow empty string

marcandre:
The check for a given `value` was oversimplified and didn't allow for `''`.

I didn't use the shorter form `null == value` because that isn't used anywhere else, but would you consider using that check (there and elsewhere) when looking for `undefined` or `null`?

From [this SO question](http://stackoverflow.com/a/21273362/8279)
> Being a more experienced JS developer now, I should note that using `if (variable == null)` is in fact the standard way to catch null and undefined simultaneously. When writing professional JS, it's taken for granted that the behavior is understood.

Also: http://stackoverflow.com/a/15992131/8279

### DIFF
--- a/src/parsley/field.js
+++ b/src/parsley/field.js
@@ -62,7 +62,8 @@ define('parsley/field', [
       if (0 === priorities.length)
         return this.validationResult = [];
       // Value could be passed as argument, needed to add more power to 'parsley:field:validate'
-      value = value || this.getValue();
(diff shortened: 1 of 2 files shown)


marcandre:
Ok, so is this good to merge as is then?

guillaumepotier:
Yes totally, sorry for the delay ;)

Best

PR #815, opened 2015-01-05
Value function

marcandre:
This PR allows `option.value` to be a function. This is not available with the `data-featherlight-value` API, of course, only using the JS api.

I realize that the same effect can be obtained by listening to the `validate` event and setting the field's `value` directly, but I feel that giving this other intuitive possibility is good.

Based on 2.1


guillaumepotier:
Hi, that's a nifty new feature, great for 2.1

Thanks!

Issue #816, opened 2015-01-05
OptionFactory refactor


marcandre:
Unless I'm mistaken, it is currently very unintuitive how to set a field option via JS after it has been initialized:

    // Either:
    $input.parsley().OptionsFactory.staticOptions.trimValue = true
    // or must be destroyed first:
    $input.parsley().destroy()
    $input.parsley({trimValue: true})

One can't simply change the options because these are reset by the `OptionFactory` by combining all different layers (e.g. global, form, field) of options.

An alternative approach to the OptionFactory approach would be to use a prototype chain instead. The global options would be the prototype of form data options, which in turn would be the prototype of the form static options, itself prototype of field data options, with the field static level being the last level.

The main advantage would be that it would allow setting options directly. For example, the following would work naturally:

    $input.parsley().options.trimValue = true

This would work as `actualizeOptions` would only change (in place) the values for the prototypes of `options`, and keep the own properties of `options` intact.

Another advantage would be that it fixes the fact that the static level (which combine default options and options passed as a parameter) is currently overridden by data options. Especially for the field level, it makes no sense to me that specific options passed via JS would be overridden by data options of the form...

I think the solution would also be quite elegant (and consume a bit less memory, but that's really negligible in the big scheme of things).

Before I work on the PR, does this sound good to you?

guillaumepotier:
Hi,

OptionsFactory is by far the worst bit of Parsley 2.x. Like you clearly stated above, this is very unintuitive, a bit overcomplicated a rework would clearly help.
I agree too that javascript options should prevail on data binded values.

I might see what you want to do here, that seems great, it may be clearer with your PR. Hope actual tests would help you do this refactoring peacefully :)

Best

marcandre:
Ok, so it wasn't too bad to write after all.

One drawback I didn't think of is that this all works beautifully, but I haven't coded for nested options. So someone doing `$field.parsley().options.validators.foo = function()` would actually modify the `ParsleyConfig.validators` object. I think `validators` is the only option that isn't flat, right?

PR #822, opened 2015-01-13
Proto options

marcandre:
This addresses #816 

Here's a summary of the commits:

#### Tweak existing tests
Improve the tests while making them compatible with what's coming. *(4 commits)*

#### Tweak existing code
Small changes to the code base but no change to any functionality or function interface *(4 commits)*

#### Use prototype for global options
Gradually modify the interface to `ParsleyUtils` to `OptionFactory` so global options are handled using `ParsleyConfig` as prototype. *(5 commits)*

#### Use protoype for field & form options
Creates `_resetOptions` that is in charge of building the options with the right prototype chain. Note that only a single test changes, because we now give priority to explicit options over DOM options.  *(1 commit)*

#### Cleanup
The only thing not handled as nicely as before is if the parent changes (only happens in one spot). Then we must rebuild the `options` (since the sequence of prototypes have changed). The last commit shows something that wasn't possible before. *(last 3 commits)*


marcandre:
For some reason, github choses to display things in [the wrong order](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/).

Please use [this list of commits](https://github.com/marcandre/Parsley.js/compare/2.1...proto_options) that shows the actual order of the commits.

guillaumepotier:
That seems a pretty neat work.

I'll try to take some time tonight to make a in-deep review. Thanks !

marcandre:
Any chance to look at it? I realize it's not a trivial PR...
Let me know if I can do anything.

src/parsley.js
```diff
@@ -212,6 +203,13 @@ define([
     }
   };
 
+  // Supplement ParsleyField and Form with ParsleyAbstract
+  // This way, the constructors will have access to those methods
+  $.extend(ParsleyField.prototype, ParsleyAbstract.prototype);
```
guillaumepotier: Nice one

src/parsley/utils.js
```diff
@@ -2,29 +2,37 @@ define('parsley/utils', function () {
...
       var
         attribute,
-        obj = {},
         msie = this.msieversion(),
         regex = new RegExp('^' + namespace, 'i');
 
+      if ('undefined' === typeof obj)
+        obj = {};
+      else {
+        // Clear all existing values. This won't affect prototype's values
```
guillaumepotier: Don't we need a `if (obj.hasOwnProperty(i))` check here ?

src/parsley/utils.js
```diff
@@ -93,6 +101,22 @@ define('parsley/utils', function () {
         .toLowerCase();
     },
 
+    objectCreate: Object.create || (function() {
```
guillaumepotier: Could benefit a little comment / doc on how this method works and what it does. Not mandatory but a plus :)

src/parsley/abstract.js
```diff
@@ -1,16 +1,30 @@
...
     actualizeOptions: function () {
-      this.options = this.OptionsFactory.get(this);
-
+      ParsleyUtils.attr(this.$element, this.options.namespace, this.domOptions);
+      if (this.parent)
+        this.parent.actualizeOptions();
       return this;
     },
 
+    _resetOptions: function(initOptions) {
```
guillaumepotier: Could benefit some comments here

src/parsley/utils.js
guillaumepotier: Found out it might be https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create ;)

test/features/abstract.js
```diff
@@ -4,14 +4,14 @@ define(function () {
...
+        expect(parsleyField.options.pattern).to.eql('[A-F][0-9]{5}');
+        expect(parsleyField.options.required).to.eql('');
 
         $('#element').removeAttr('data-parsley-pattern');
         parsleyField.actualizeOptions();
 
-        expect(parsleyField.options).not.to.have.key('pattern');
-        expect(parsleyField.options).have.key('required');
+        expect(parsleyField.options.pattern).to.be(undefined);
+        expect(parsleyField.options.required).to.eql('');
```
guillaumepotier: Since `_resetOptions` is used by various objects (even if `_` means this is a private method and should not be called in public api) won't it be good to add some tests here or tests for field / form are sufficient?

src/parsley/utils.js
marcandre: Right, prototype's values are not affected by `delete`, but just for clarity it's probably better to add it. Done.

src/parsley/utils.js
marcandre: Indeed, that's where I took it from. Comment added.

test/features/abstract.js
marcandre: I think that the current tests cover the trickiness of options pretty well. I test private methods anyways (and I notice that `_bindConstraints` and `_bindHtml5Constraints`, which are more complex, aren't tested either).

guillaumepotier:
This is a truly great PR. Thanks again for your awesome work.

PR #825, opened 2015-01-16
Check for empty value

orx0r:
If related field is empty then we do not compare.


marcandre:
I'm afraid this PR is not acceptable without test.
Moreover, I'm not sure what you are trying to achieve here. An empty string is not greater or smaller than any number, no?

marcandre:
Closing, will reopen if we get feedback

PR #832, opened 2015-01-22
Update parsley.css

coliff:
Remove unneeded vendor prefixes for opacity. The Moz-opacity was only ever needed for before Firefox 1.0 and I don't think webkit-opacity was ever used at all.
REF: http://css-tricks.com/css-transparency-settings-for-all-broswers/
REF: http://caniuse.com/#feat=css-opacity  (click show all - no vendor prefixes on record)


guillaumepotier:
Thanks

Issue #813, opened 2015-01-02
Remote parsley should have option to disable caching


jbojcic1:
Currently parsley remote caches every response and that can be a problem. For example I am working on a single page app and when creating user I am checking is email unique so this happens:
1. I normally create user with email which wasn't used earlier (parsley response is cached)
2. I try to create user with same email again and since response is cached it passes validation again

marcandre:
Probably the best solution is for the cache to be owned by the *form* instead of the *validator*. This would resolve your issue, right?

guillaumepotier:
Hi @jbojcic1,

I'm using parsley remote @Wisembly for the same purpose and it works well! Like @marcandre said this is not Parsley role to call api to create objects, since it only send (normally) only the input value, not the whole form.

Moreover, the parsley remote url should make a `GET` call while the form make a `POST` one to create the email in your database.

In parsley point of view (following REST principle) every unique url call should always return same response and do the same action. That's why we cache every request that is made to avoid too much stress on servers especially for existence checking like you do in your case.

I might add a parsley remote example in documentation, this your use case is quite typical and should not confuse users.

Best

marcandre:
I presume that @jbojcic1 is using ajax to `POST` the user creation, so the parsley lib is never reloaded, while @guillaumepotier probably uses the default form submission which reloads the whole page?

> every unique url call should always return same response and do the same action

Well, no and yes. A `GET` should be *idempotent*, i.e. not have side effects server-side. The response can be different every time though. Otherwise we'd be reading the same news everyday :smile:

I think we should probably invalidate caches just before form submission (onSubmitValidate, and in the remote case just before we re-submit the form). @guillaumepotier can you think of a case when that's would create issues?

guillaumepotier:
We have to make sure it solves @jbojcic1 case but no that seems a pretty good idea yes.

jbojcic1:
Yes, like @marcandre said, I use POST to create user not form submission.

jbojcic1:
@guillaumepotier @marcandre 
I never submit my form in a

guillaumepotier:
Like @marcandre said

> I think we should probably invalidate caches just before form submission (onSubmitValidate, and in the remote case just before we re-submit the form). @guillaumepotier can you think of a case when that's would create issues?

we might even clear cache automatically after a validation success, what do you think, would it solve your problem?

jbojcic1:
[...]

PR #842, opened 2015-01-31
Clear cache

marcandre:
I think this solves #813 and adds a useful event `form:submit.parsley`.

@guillaumepotier what do you think?


guillaumepotier:
Hi,

Yes that seems great, especially the interruptible submit event!

Good to me

Issue #710, opened 2014-08-14
Parsley trigger stops working after calling parsley.reset()


marioBonales:
It appears that after calling parsley().reset() the data-trigger gets debinded and it no longer validates on the specified trigger:

Here's a js-fiddle: http://jsfiddle.net/67sj6o1k/1/

Steps to reproduce:
1. Write a non-url in the input, parsley works great validating on keyup
2. Click the parsley reset button
3. Write a non-url in the input again, this time it doesn't validate on keyup

Expected: Parsley should work on the specified trigger after reset

jfitzell:
Me too

marcandre:
I have the impression that `reset` is now obsolete. Are there any legitimate uses anymore?

jfitzell:
Not sure... my use case is that I'm doing a multi-stage form. I'm validating a subset of the form for each "page" but if the user goes back to the previous page I don't want to show all the previously validated fields highlighted. So I'm resetting the form every time I change page. A bit ugly, so happy for another suggestion. What has made reset() obsolete?

I've also just written tests that show the bug and found a fix - just writing up a commit message and pull request so I can submit.

jfitzell:
Pull request in #843

marcandre:
My bad, got confused with `destroy`.
Thanks so much for the PR!

PR #843, opened 2015-02-04
Re-register failing field trigger after reset() [#710]

jfitzell:
Fix issue where failing fields were not revalidated after resetting:
- Reset _ui.failedOnce to false in reset()
- Relax check in reflow() to allow undefined or false


marcandre:
Thank you for this great PR :smile:

PR #844, opened 2015-02-04
Explicit overriding of an element's key by using a data attribute

danielgreane:
There are situations where the name of an input (used for form submission) does not match that of the remote validation service.

For example, the local service accepting the form submission may expect the name `emailAddress`
A 3rd party validation service, such as postcode anywhere, is expecting an `Email` field

This addition allows you to either set the 3rd party field name before addAsyncValidator is called (using $.data), or by directly attaching a data-name attribute on the element

### DIFF
--- a/src/extra/plugin/remote.js
+++ b/src/extra/plugin/remote.js
@@ -184,7 +184,7 @@ window.ParsleyExtend = $.extend(window.ParsleyExtend, {
       throw new Error('Calling an undefined async validator: `' + validator + '`');
 
     // Fill data with current value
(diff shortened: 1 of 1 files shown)


marcandre:
Thanks for the PR.

I understand you want to customize the ajax calls before it is send over (in your case to modify the keys). This is not the first request we have to modify the ajax call, so maybe the best way to handle this would be have a more generic way to prepare/send the ajax call.

Even if the reason to customize ajax calls for most people was to change keys, the solution you are proposing would be better handled using a dedicated `data-parsley-something` option. Finally, there are no tests.

PR #852, opened 2015-02-25
Fix jshint error

vsn4ik:
http://api.jquery.com/jquery.noop/


vsn4ik:
Travis CI complete the task without errors.

marcandre:
The whitespace and semicolons could be ok, the `$.noop` is not acceptable as we then go on modifying it's prototype. I don't want to mess with `$.noop`!

In any case, the fact that `ParsleyAbstract`'s construction is empty is true for now, could be changed in the future.

vsn4ik:
`$.noop` removed.

marcandre:
Awesome, thanks for the PR :+1:

We should include that in the test suite.

PR #853, opened 2015-02-26
Add an option to "squish" whitespace

shaneog:
The squish-value option squishes _all_ whitespace in the value, by
replacing multiple sequential whitespace characters with a single
space.
It also trims the value.

This can be useful when someone is attempting to get around min-length
validators by using strings like `a<multiple spaces>b`.

_Edit: Ironically I can't show a proper example here since GitHub squish 
the text in this field_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Parsley 2.x changelog
 
 ## master
-
+  
+- added `data-parsley-squish-value` option
 - Support for validators with compound names by restoring ful case sensitivity
   to error messages. (#805)
 - new event 'form:submit.parsley' fired before a form is submitted.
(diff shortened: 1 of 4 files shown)


shaneog:
Updated PR based on comments.

marcandre:
Thanks :smile:

I'm fine with the PR, well written too.

(cc @guillaumepotier) I'm just wondering if it wouldn't be best to have a api `whitespace=trim` and `whitespace=squish` instead of `trim=true` and `squish=true` (even though we'd have to keep `trim=true` for a while for compatibility). Thoughts?

shaneog:
Agreed that a `whitespace` API would be much more efficient now that there are two options.

shaneog:
I created a new PR with a `whitespace` API, so feel free to comment there :smile:

Issue #764, opened 2014-11-12
Parsley's pattern attribute handling doesn't match the HTML5 spec


liamjones:
The HTML5 spec states the following for the [pattern attribute](http://www.w3.org/TR/html/forms.html#attr-input-pattern):

> The compiled pattern regular expression, when matched against a string, must have its start anchored to the start of the string and its end anchored to the end of the string.
>
> **Note:** This implies that the regular expression language used for this attribute is the same as that used in JavaScript, except that the pattern attribute is matched against the entire value, not just any subset (somewhat as if it implied a ``^(?:`` at the start of the pattern and a ``)$`` at the end).

Parsley doesn't anchor the regexp it reads out of a pattern attribute so patterns which would result in a failure with raw HTML5 browser validation start allowing values through once Parsley is included.

See test case: http://jsfiddle.net/9kg8g9w0/3/

If Parsley only used it's own attributes (data-parsley-pattern) then I think this differing behaviour would be fine. Since it's pulling from HTML5 validation properties I'd personally expect its behaviour to match that of a HTML5 validation on the same property.

marcandre:
I agree, at least for "standard" regex.
Parsley also extends regex by accepting a full pattern (e.g. "/hello/i"). These could remain unanchored?
v2.1/3.0 would be a good occasion for this.

liamjones:
Agreed, there should probably be some way of specifying a full pattern as well. I'm wondering if it should only be allowed via a parsley specific attribute though?

My thinking; if it's specified in the pattern attribute and the page ends up falling back to HTML5 validation (say, Parsley doesn't load or JS is disabled) then the validation is going to behave differently. E.g. `/.{0.5}/i` in the pattern attribute would look for the literal `/` and `i` characters within your input (and it's still anchored automatically by the browser too).

marcandre:
It's true that we're "incompatible" if pattern is set to `/.{0.5}/i` as this will be interpreted literally in HTML5. On the other hand, I just can't see anyone using such a pattern for real. If they really need that (say they want someone to enter a regex?!?), they can escape the `/`, like `\/.{0.5}/i` would be the same in parsley and html5.
In short, I can sleep peacefully with such an "incompatibility", it should just be noted in the specs.
The main question though is this: should "extended" regex be anchored too or not?

marcandre:
@guillaumepotier Any opinion on the above?

guillaumepotier:
[...]

PR #854, opened 2015-02-26
Add new whitespace API

shaneog:
The new `whitespace` API handles actions on whitespace, which at the
moment consist of `trim` and `squish`.
- `trim`: removes leading and trailing whitespace characters from the
value.
- `squish`: replaces multiple sequential whitespace characters with a
single space, and then performs a `trim`.

This API can be used by the following data attribute-value pairs:
`data-parsley-whitespace=“trim”` or `data-parsley-whitespace=“squish”`

This PR deprecates the use of `data-parsley-trim-value=“true”`.

This PR replaces #853.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master
 
+- Deprecated `data-parsley-trim-value` in favour of new `whitespace` API
+- Added `whitespace` API with two options: `trim` and `squish`
+
 ## 2.1.2
 
 - fix custom triggers after a `reset()` (#926)
--- a/doc/index.html
+++ b/doc/index.html
@@ -412,8 +412,8 @@ if (false === instance.options.priorityEnabled)
                 <td>A field is by default not validated if it is not required and empty. By adding <code>data-parsley-validate-if-empty</code>, validation will be done even if field is empty. Useful if you need some custom validators that check something particular when a field is empty.</td>
               </tr>
               <tr>
-                <td><code>data-parsley-trim-value</code> <version>#2.0</version></td>
-                <td>Trim field value <strong>only for Parsley validation</strong> (and not inside the input itself, data sent by your form won't be trimmed). Useful if your backend already does so and if trailing spaces could unnecessarily mess with your validation. Use: <code>data-parsley-trim-value="true"</code>.</td>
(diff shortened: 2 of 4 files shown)


marcandre:
Thanks for updating your PR with the suggested api.
I think it would be worth issuing a deprecating warning if someone uses `trim-value`.
I'm not super found of the current `_whitespace` helper method; either it should not exist, or it should read and use the `this.options.whatever`, since it's a method (and in that case something like `_handleWhitespace` would be a better name).
I'm not sure `getValue` should have that many `return` statements. I mean, if we have another option that, for example, would camelCase a string, say, then it would need to be compoundable with `whitespace` option.
Anyways, you seem to be eager to tweak your PR(s), but if you're not I could do it too :-)

shaneog:
Good points. I've updated the PR now.

I'm not sure how to deal with `getValue` and multiple value actions, so I've avoided it in this PR.

shaneog:
Fixed the build - it was only failing on Travis CI and not locally. Travis CI was considering the deprecation warning as an error :confused:

marcandre:
Thanks for the update. I made some comments, assuming you're ok with doing the tweaks...

shaneog:
I've made the changes but I'm having a little trouble with the tests and `expectWarning`. For some reason it's not working...
Could you point me in the correct direction @marcandre please?

marcandre:
Ok, I fixed the test suite, so `warnOnce` should be testable "naturally". This made me realize there were a bunch of tests still using the old pubsub api.
@guillaumepotier Are you :+1: with this api? I think + #764 would make a good v2.1.

shaneog:
Great, now it's using `warnOnce` :+1:

marcandre:
Great, thanks for your work :+1:

I've assigned this to @guillaumepotier for his approval. He's in vacation for a month, so don't expect feedback for a while though.

shaneog:
Updated PR so it merges cleanly

guillaumepotier:
This look awesome to me :) Sorry for reviewing that so late, very busy atm :(

Good for me @marcandre (and I'll try to have a look at yours during this week end too)

Best

shaneog:
Updated PR so that it merges cleanly

marcandre:
Awesome, thanks!

PR #857, opened 2015-03-04
Fixed a bug where specifying a minlength and maxlength of the same value...

mmcev106:
...did not work properly (only when use the HTML5 versions of those options).

Don't hesitate to ask if you'd prefer that I solve this in a different way.


marcandre:
Thanks for the PR :smile: 

I haven't looked closely at the code, but I'm wondering if the parsing shouldn't be downstream, i.e. at the validator level instead.

mmcev106:
I think you're right.  I'll close this request and make a new one for validator.js.

PR #861, opened 2015-03-05
Anchored pattern

marcandre:
We should anchor `<input pattern="foo">`, to be compatible with HTML5.
We can definitely accept `<input data-parsley-pattern="/foo/i">` as being unanchored.

The question is: what to do with `<input data-parsley-pattern="foo">` (anchored or not?) and `<input pattern="/foo/i">` (unanchored regex or anchored pattern starting with a slash?)

This PR suggests that `pattern` or `data-parsley-pattern` are dealt with the same way, so that patterns of the form `/<anything>/{flag}` are interpreted differently from HTML5 as litteral regex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.2.0-rc2
 
 - Remote: use HTTP status code for what is considered valid or not (#956)
+- pattern validator is now anchored, unless it looks like /pattern/flag (#861)
 
 ## 2.2.0-rc1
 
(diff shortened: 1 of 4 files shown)


marcandre:
@guillaumepotier We could really go either way on this, I'd like your decision on this...

marcandre:
PS: This would fix #764

marcandre:
Updated this from `master`.
@guillaumepotier ping!

guillaumepotier:
Seems good to me!

marcandre:
Merge, thanks

PR #879, opened 2015-03-31
Custom validator messages with compound names (via validator#addMessage)

alampros:
Add tests for custom error messages without using data attributes (using validator#addMessage).

May be related to #806

The bug can be seen in this [jsfiddle](http://jsfiddle.net/alampros/dknc3nfk/).

### DIFF
--- a/src/parsley/validator.js
+++ b/src/parsley/validator.js
@@ -54,7 +54,7 @@ define('parsley/validator', [
       if ('undefined' === typeof this.catalog[locale])
         this.catalog[locale] = {};
 
-      this.catalog[locale][name.toLowerCase()] = message;
+      this.catalog[locale][name] = message;
 
       return this;
     },
--- a/test/features/ui.js
+++ b/test/features/ui.js
@@ -333,6 +333,33 @@ define(function () {
         window.ParsleyValidator.removeValidator('customValidator');
       });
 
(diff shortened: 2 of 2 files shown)


marcandre:
Awesome PR, thanks so much!

What would you think of DRYing up a bit the new tests, either combining them or putting the common setup in a function?

alampros:
Sure thing. I can just remove some repeated setup.

marcandre:
Mmm, thanks for modifying your PR, but I feel tests should be independent, it's a really bad idea to rely on side effects from another test.

alampros:
You're quite welcome, and I couldn't agree more. This feels like it should be done in setup/teardown (`before()` and `after()` in should.js terms) functions, but that would mean putting these tests in their own suite, right? Perhaps a nested suite?

marcandre:
I think you can nest tests (with their setup/teardown).

I'm also fine if you want to test both as in the same test, they're really about the same thing.

alampros:
Anything else I can do on this?

marcandre:
Oh, sorry, didn't see you had made the changes.

Looking good.

I think if you remove the references to `done` and squash the commits, we'll be good to go :smile:

marcandre:
All fixed now. Thanks!

PR #880, opened 2015-03-31
Example of a custom validator with compound name

alampros:
Adds a bit of clutter, but it might help people understand the data-attribute to camelCase relationship a bit easier.


marcandre:
That's a good point. Maybe we can simply change `multipleof` with `multiple-of`? No clutter, and we give the idea

marcandre:
Doc tweaked. Example also has 3 such examples. Thanks for the suggestion.

Issue #888, opened 2015-04-15
checkbox with the same name validate in different form


ChJJin:
Hi, 

I have two form on the same page. There are more than one checkbox with the same name in both of the forms. 

I have to validates the range of checked checkboxes in each form, with ```data-parsley-check```. But in fact, when I check different checkbox with same name in different form, it validates the numbers by adding the checked checkbox in the both form together !! And the error messages are only listed in the first form !!

It shows that the checkbox with the same name is validated on the whole page.

marcandre:
Agreed, that's a bug. Shouldn't be too hard to fix. You want to try? :smile_cat:

ChJJin:
I'd like to try, it's my pleasure, thx.  :smile_cat:

marcandre:
Great, feel free to ask any question. Be careful that it is possible that some inputs might not be inside a `<form>`, in which case we need to consider the `<body>` as the "container"

ChJJin:
Got it .
These situations does not come to my mind just now, thx. :+1:

marcandre:
Good start btw.
Also, you will need to add some tests that fail before and work after your changes.

marcandre:
I have a solution, currently in a branch.

marcandre:
Merge and fixed a while ago. Thanks for reporting this.

PR #889, opened 2015-04-15
fix the bug that checkbox/radio validate in whole page

ChJJin:
#888


marcandre:
Yeah, I've actually been contemplating `parent` pointing to either an instance of `ParsleyForm` or at `Parsley` itself, and `Parsley.$element` being the document/body.
Not yet though.

claudiomedina:
Any evolution on this problem?

Issue #627, opened 2014-04-30
remote validation limited support on rest style URL


xfcjscn:
For rest style URLs, normally they are the combination of baseURL and resource ID.
E.g. http://xxx.com/books/1
And the resource ID normally populated with the input's value. 
In this case, an interface to manipulate the URL is required. But currently this is not supported in remote validator. Can this be added?

marcandre:
This is now easy to do in `master`, with `url: "http://api/checkvatid/{value}"`

PR #894, opened 2015-04-20
added ability to change remote options on the fly

DiliBau:
When trying to validated a field together with an adjacent field, the `parsley:field:validate` is simply not powerful enough to allow you to send additional data dynamically and neither is the `data-parsley-remote-options` attribute.

This applies, for example, when remotely validating international phone numbers which require the user to also select the country in order to apply formatting rules.

```javascript
// this fails and data is never sent
$.listen('parsley:field:validate', function (field) {
  if (field === someRefToMyField) {
    field.options.remoteOptions.data = {someName: $someField.val()};
  }
});
// this works
$.listen('parsley:remote:options', function (field, options) {
  if (field === someRefToMyField) {
    options.data = {someName: $someField.val()};
  }
});
```

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -1103,6 +1103,47 @@ $('[name="q"]').parsley()
           </ul>
         </p>
 
+        <h3 id="remote-events-list">Events List</h3>
+        <table class="table table-stripped table-bordered">
+          <thead>
+            <tr>
+              <th class="col-md-2">Name</th>
+              <th class="col-md-1">Arguments</th>
+              <th class="col-md-1">Fired by</th>
+              <th class="col-md-4">Description</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td><code>field:validate.parsley</code> <version>#2.1</version></td>
+              <td><code>event</code>, <code>ParsleyField</code>, <code>xhrOptions</code></td>
+              <td><code>.asyncIsValid()</code></td>
+              <td>
+                Trigger <strong>after</strong> the raw XmlHttpRequest options are built, <strong>before</strong> the actual call to <strong>$.ajax()</strong>.
+                For further information about available options and setting see <a href="http://api.jquery.com/jquery.ajax/#jQuery-ajax-settings">$.ajax()</a>.
+              </td>
+            </tr>
+          </tbody>
+        </table>
+
+        <h3 id="remote-events-usage">Events Usage</h3>
+        <p>Configuring the listener is really straight forward:
+<pre><code>&lt;input id="q" name="q" type="text" data-parsley-remote data-parsley-remote-validator='mycustom' value="foo" />
+[...]
+&lt;script href="parsley.remote.js">&lt;/script>
+&lt;script href="parsley.js">&lt;/script>
+&lt;script type="text/javascript">
+$("#q").on('field:ajaxOptions.parsley', function (evt, field, xhrOptions) {
+  xhrOptions.url = 'http://new.url';
+  if (!('data' in xhrOptions)) {
+    xhrOptions.data = {};
+  }
+  xhrOptions.data['new-parameter'] = 'new-value';
+});
+&lt;/script>
+</code></pre>
+        </p>
+
         <!-- ****************** Extra ****************** -->
         <h1 id="extras" class="page-header">Parsley Extras</h1>
         <p>
@@ -1300,6 +1341,10 @@ $('[name="q"]').parsley()
               </li>
               <li>
                 <a href="#remote">Parsley Remote</a>
+                <ul class="nav">
+                    <li><a href="#remote-events-list">Events</a></li>
+                    <li><a href="#remote-events-usage">Usage</a></li>
+                </ul>
               </li>
               <li>
                 <a href="#extras">Parsley Extras</a>
--- a/src/extra/plugin/remote.js
+++ b/src/extra/plugin/remote.js
@@ -198,6 +198,9 @@ window.ParsleyExtend = $.extend(window.ParsleyExtend, {
       type: 'GET'
     }, this.options.remoteOptions || {});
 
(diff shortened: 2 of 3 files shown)


marcandre:
Thanks for the PR.

An easier approach might be to allow `remoteOptions` to be a function, to be called before an ajax call. There are other requests like #627 to tweak the URL, so the solution needs some reflexion.

DiliBau:
@marcandre since the hash that is passed to this callback is the options object that is sent to the `$.ajax()` call, the `url` could easily be changed by doing something like
```javascript
$.listen('parsley:remote:options', function (field, options) {
  if (field === someRefToMyField) {
    options.url = field.$element.data().validationUrl || someFunctionToGenerateTheURL();
  }
});
```

guillaumepotier:
I kinda like this approach since it is very lightweight in term or code and allow various tweaks. This seems a bit more like a 'hack' or a 'patch' and as mentioned in #627 proper 'native' implementation should be addressed, meantime this solution could do the job IMHO.

This PR though lacks some documentation and tests :)

Best

DiliBau:
@guillaumepotier will add tests and docs over the course of today

marcandre:
If we're going to add another event for this, I'd call it `'parsley:field:something'` since it's triggered on a field. Maybe `'parsley:field:ajaxoptions'`?

Still, I'm wondering when someone would ever want to bind more than once to that event. If we never bind more than once, it should be an option instead, no?

guillaumepotier:
Let's say you have a pseudo field that check the pseudo existence in database (like Twitter username on signup), using a REST API, successive calls could be

```
GET api.yourservice.com/usernames/foo
GET api.yourservice.com/usernames/foobar
GET api.yourservice.com/usernames/foobarbaz
```

Could be useful to modify on the fly the url on each keypress, no ?

marcandre:
Definitely useful!

My question is: when would a given field's ajax options be modified via `'parsley:field:ajaxoptions'` by more than 1 listener? If always 0 or 1 listener, pubsub is overkill and an option `ajaxOptions` would be fine. If sometimes more than 1 listener, then event is much better, as otherwise the inner option `ajaxOptions` has to be aware of the outer one.

guillaumepotier:
Ok I see the point now. How could we maybe work that way @DiliBau ?

marcandre:
Mmm, I guess a form listener could change the URL in a RESTful way and a local listener could tweak the values according to some kind of formula, so after some more consideration, I'm siding with events after all :smile:

So yeah, the more I think about it, the more awesome that solution sounds. “Only fools never change their mind.”

DiliBau:
@guillaumepotier sorry for the delay, i've updated the PR with docs, examples and a test

guillaumepotier:
Thx @DiliBau!

Unfortunately, it seems there is a problem with the test suite.

```
browserify test/env/browserify.js > test/env/browserify_bundle.js
Error: Cannot find module 'jquery' from '/home/travis/build/guillaumepotier/Parsley.js/dist'
    at /home/travis/.nvm/v0.10.38/lib/node_modules/browserify/node_modules/resolve/lib/async.js:46:17
    at process (/home/travis/.nvm/v0.10.38/lib/node_modules/browserify/node_modules/resolve/lib/async.js:173:43)
    at ondir (/home/travis/.nvm/v0.10.38/lib/node_modules/browserify/node_modules/resolve/lib/async.js:188:17)
    at load (/home/travis/.nvm/v0.10.38/lib/node_modules/browserify/node_modules/resolve/lib/async.js:69:43)
    at onex (/home/travis/.nvm/v0.10.38/lib/node_modules/browserify/node_modules/resolve/lib/async.js:92:31)
    at /home/travis/.nvm/v0.10.38/lib/node_modules/browserify/node_modules/resolve/lib/async.js:22:47
    at Object.oncomplete (fs.js:108:15)
```

marcandre:
May be due to recent change of jQuery path: 4067c4b5ff5b16e6

marcandre:
Test looks great!

Couple of things I'd change:
- `parsley:remote:options` is not the right name (as I mentionned before), it should start with `parsley:field:`. I recommend `parsley:field:ajaxoptions`, or maybe even `parsley:field:ajax` since we could provide the possibility to the user to make the ajax call itself and return the promise?
- the PR shouldn't make any changes to `dist/`, that happens only when we make a release
- triviality, but it took me a minute to notice the difference between the two urls in the test, would be best to make them more different than two letters being swapped.

marcandre:
Couple more things:
* would be best to remove the custom validator after, to avoid potential interactions with other tests
* in the test, the listener should be added to the `parsleyInstance`, not as a global listener.
* it's not obvious why the validator fails, better `return false`, or why not return `true` and use `done`?

DiliBau:
@marcandre @guillaumepotier 
- renamed event to `field:remote.ajaxOptions`  to be consistent with the other events
- reverted changes to `dist`
- updated events to use `on()`, `off()` instead of global `$.listen()`
- updated docs
- fixed the test
- using done to avoid this
```
timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

marcandre:
Good work!
I'm confused though: did you remove your `mycustom` validator somehow? I just meant you should call `removeAsynchValidator`. But it looks like that method is missing! So simply use a very specific validator name, we might get around fixing that at some point.

More nitpicking: for reasons I can't fathom, JS events seem to be all lowercase, e.g. `dragend`, `gotpointercapture`, so I wonder if we should follow the (ugly) convention and use `ajaxoptions` instead of `ajaxOptions`.

DiliBau:
@guillaumepotier if you want the code, you can find it in my fork, you're free to take it and include it in the main report and/or modify it as you see fit
the example that i included is clear enough to get you started and the tests are passing, other then that i do not have enough time to make @marcandre happy

marcandre:
This is now merged in `master`. There's also a simpler way to use the input value in the URL

PR #899, opened 2015-04-27
Change event API again.

marcandre:
@guillaumepotier, do you have a second to look at this? It changes the API for events and ditches jQuery events. On Chrome it almost halves the binding time on an example with lots of radios.

This is not finished yet, but a good step in the right direction.

Note: `parsleyInstance.parent` is no longer `undefined` for forms / fields, it now points to `Parsley`. Basically it mimicks the DOM: input's parent is form, form's parent is Parsley (document), and Parsley has no parent. I'd like to have `Parsley.$element` == `$(document)` (or `body` or `window`), not sure, to better mimick that.

Not sure how strange or intuitive this proposal is to you, let me know...


doc/index.html
```diff
@@ -242,16 +232,70 @@ <h3 id="usage-overview">Overview</h3>
             </p>
           </div>
 
+          <h3 id="usage-configuration">Configuration</h3>
```
guillaumepotier: Nice

src/extra/plugin/remote.js
```diff
@@ -264,9 +264,8 @@ window.ParsleyConfig.validators.remote = {
   priority: -1
 };
 
-$(document).on('form:submit.parsley', function (evt, parsleyForm) {
-  if (!evt.isDefaultPrevented())
-    parsleyForm._remoteCache = {};
+window.Parsley.on('form:submit', function () {
+  this._remoteCache = {};
```
guillaumepotier: I think that if you remove cache but form submit is prevented because invalid, it could lack for future calls.
But was more a nice to have than an absolute requirement (dunno if it really occurs)

src/parsley.js
```diff
@@ -23,174 +23,28 @@ define([
   'parsley/field',
   // `Multiple` Class. Extend `ParsleyField` to generate `ParsleyFieldMultiple`
   'parsley/multiple',
+  // Factor to create Parsley instances (Form, Field or FieldMultiple)
```
guillaumepotier: Factor?

src/parsley.js
```diff
@@ -212,7 +66,7 @@ define([
       return;
     }
 
-    return new Parsley(this, options);
+    return new ParsleyFactory(this, options);
```
guillaumepotier: SO MUCH READABLE HERE!!! <3

guillaumepotier: :beers:

src/parsley.js
```diff
@@ -220,9 +74,15 @@ define([
   if ('undefined' === typeof window.ParsleyExtend)
     window.ParsleyExtend = {};
 
-  // ### ParsleyConfig
+  // ### Parsley config
   // Inherit from ParsleyDefault, and copy over any existing values
-  window.ParsleyConfig = $.extend(ParsleyUtils.objectCreate(ParsleyDefaults), window.ParsleyConfig);
+  Parsley.options = $.extend(ParsleyUtils.objectCreate(ParsleyDefaults), window.ParsleyConfig);
+  window.ParsleyConfig = Parsley.options; // Old way of accessing global options
```
guillaumepotier: Thks for the BC Breaks ;)

src/parsley/abstract.js
```diff
@@ -8,14 +8,13 @@ define('parsley/abstract', [
...
-      if (this.parent)
+      if (this.parent && this.parent.actualizeOptions)
         this.parent.actualizeOptions();
       return this;
     },
 
     _resetOptions: function (initOptions) {
-      var baseOptions = this.parent ? this.parent.options : window.ParsleyConfig;
-      this.domOptions = ParsleyUtils.objectCreate(baseOptions);
+      this.domOptions = ParsleyUtils.objectCreate(this.parent.options);
```
guillaumepotier: great

src/parsley/factory.js
```diff
@@ -0,0 +1,165 @@
+define([
```
guillaumepotier: This file is much better here but like we discussed still a mess --'

src/parsley/field.js
```diff
@@ -83,13 +83,12 @@ define('parsley/field', [
...
 
       // If we want to validate field against all constraints, just call Validator and let it do the job
-      if (false === this.options.priorityEnabled)
-        return true === (this.validationResult = this.validateThroughValidator(value, this.constraints, 'Any'));
-
-      // Sort priorities to validate more important first
-      var priorities = this._getConstraintsSortedPriorities();
-
-      // Else, iterate over priorities one by one, and validate related asserts one by one
+      var priorities = ['Any'];
```
guillaumepotier: ok why not

src/parsley/pubsub.js
```diff
@@ -42,7 +42,7 @@ define('parsley/pubsub', [
     if ('function' !== typeof arguments[1])
       throw new Error('Wrong parameters');
 
-    $(document).on(eventName(name), adapt(callback, context));
+    window.Parsley.on(eventName(name), adapt(callback, context));
```
guillaumepotier: Like we discussed too, `window.Parsley` <=> `document` for events seems right

guillaumepotier:
As usual this is a pretty neat PR ;) Once you're happy with the events API, don't hesitate to merge. We'll try to package soon the next release.

Best

src/extra/plugin/remote.js
marcandre: Right, but with our own API, the way to "prevent the default" is to interrupt the events, there's no other way, in which case we're not even called.

marcandre:
Should be good to go.
I'll merge and make a new rc

marcandre:
Built [2.1.0-rc9](https://github.com/guillaumepotier/Parsley.js/releases/tag/2.1.0-rc9)

PR #901, opened 2015-04-29
Add Portuguese (Portugal) translation.

jcalado:
Adds the i18n file for a Portuguese translation.


marcandre:
Looks good, but shouldn't the key be `pt` instead of `pt-pt`?

marcandre:
Ping @jcalado ?

marcandre:
Oh, I see, there's a 'pt-br' translation already. Deifnitely not sure we need two translations for this, but merged anyways. Thanks

claudiomedina:
pt-br is for brasilian portuguese (https://en.wikipedia.org/wiki/Brazilian_Portuguese) that has some different terms.

Issue #905, opened 2015-05-07
Multiple asynchronous validation checks


Flaburgan:
Hi everyone.

I'd like to validate that an email is correct when the user registers. For that, I need to:
1. Check that the email is valid (regex but also lookup on the domain to see if there is an MX, so this has to be done server side)
2. Check that the email is available, that means it is not already in my database

So I had a look at [parsley remote](http://parsleyjs.org/doc/index.html#remote) and didn't like the Parsley way to check the xhr status with 200 or 404. My API is here, it should never respond with 404. So I checked  how to write a custom validator. I have two things to check, but `data-parsley-remote-validator='my-custom'` allows me to define only one API call (and also one error message).

It looks like the easiest way to deal with this problem would be to implement https://github.com/guillaumepotier/Parsley.js/issues/560 == to allow to get an error message from the server instead of a simple boolean. That way, I could answer "Email invalid", "Email unavailable", or "" which would mean no error. But honestly, that feels kind of a hack. So my question is, why can't we go with `data-parsley-my-custom=''` (or `data-parsley-remote-my-custom=''`) instead of `data-parsley-remote-validator='my-custom'`? This syntax would be similar that for other personalized (synchronous) validators. Is there any reason this has not be done that way?

marcandre:
I'd write two separate validators. For one thing, the options for the domain name should only contain the domain name so it gets cached clientside.

Not a direct answer, but there are other acceptable status codes beside 404, in particular [403 "Forbidden"](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.4) which I feel would be quite meaningful in the context of a parsley remote validation.

marcandre:
Please check the latest release.

In addition to the classic `remote` validator, it's now easy to write your own validator returning a promise

PR #909, opened 2015-05-11
Asyncmethod

pascalheraud:
Hello,

We have multiple issues regarding asyncMethod : 
- We want to use more than one async validator per field
- We want different messages for each validator
- We need to make the XHR call ourself (because we rely on a layer to do that).

The current implementation of async if too much based on jquery XHR and we propose to add a new async validator type.
The current one is the "xhr" type because it's base on XHR and we added a "method" one.
The goal of this validator is to provide only a method with a callback.
The method to async stuff and calls the callback with a boolean to say if it's success or not.

To use it, one has just to add a "addAsyncMethod".
I've modified the dev.html file for having a clear example of how to use it. I don't think it has to be merged.

I can provide tests and docs if you plan to merge it.

Pascal / La Roue Verte.

Issues related : 
https://github.com/guillaumepotier/Parsley.js/issues/905

### DIFF
--- a/src/extra/plugin/remote.js
+++ b/src/extra/plugin/remote.js
@@ -12,8 +12,12 @@ window.ParsleyExtend = $.extend(window.ParsleyExtend, {
 
   asyncValidators: $.extend({
     'default': {
+      fntype:'xhr',
       fn: function (xhr) {
         return 'resolved' === xhr.state();
+      },
+      asyncmethod: function(value, callbackboolean){
+    	  callbackboolean(true);
       },
       url: false
     },
@@ -35,7 +39,19 @@ window.ParsleyExtend = $.extend(window.ParsleyExtend, {
 
     return this;
   },
+	  
+  addAsyncMethodValidator: function (name, fn, url, options) {
+    this.asyncValidators[name.toLowerCase()] = {
+      fn: fn,
+      fntype:'method',
+      asyncmethod: fn,
+      url: url || false,
+      options: options || {}
+    };
 
+    return this;
+  },
+  
   asyncValidate: function () {
     if ('ParsleyForm' === this.__class__)
       return this._asyncValidateForm.apply(this, arguments);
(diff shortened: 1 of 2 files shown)


pascalheraud:
By the way, the proposed the code is only here to show what could add for solving our issue.
If you agree with the concept, I will provide a better code (some arguments from addAsyncMethodValidator are not used for example), tests and doc.

marcandre:
I'm not sure I understand how the first two points are a problem currently.

As for the third point, I think a nice way would be to allow validators (`addValidator`, not `addAsynchValidator`) to return promises and deal with those accordingly. That would resolve your issue, right?

Flaburgan:
@marcandre well if the first point is not a problem, what's the syntax to have two async validation on a field?

pascalheraud:
@marcandre validator returning Promises would be nice indeed. I think it will achieve the same goal.

marcandre:
My bad, one remote only, so first point quite valid. My proposed solution of simply handling promises would resolve that, although one has to make his own ajax calls

Flaburgan:
We are Java developers, I have no idea what a promise is. So not sure we can write this code. Is this something you would like to implement?

marcandre:
I definitely would like to write this, but I have other things to write too...

anedisi:
it would be also great to have the ability to get related filelds in validator (remote and local)
sometimes you have "select" field on what the validation of my other field depends.
i would like the ability to define related fields and that i can send them with that field to the server
right now im changing the the options.data on the fly but it would be great to have some "native" way to do this.

marcandre:
The latest release 2.2 makes it very easy to make your own custom validators returning promises (you can even return the `xhr` directly).

Let me know if it doesn't completely solve these issues.

Thanks!

Flaburgan:
@pascalheraud and I are going to test it, thank you!

PR #920, opened 2015-06-08
Update luhn extra test to work with 2.0

johanwikstrom:


marcandre:
I think the original formula is intended to be loadable before or after Parsley (I don't particularly like that "feature")

marcandre:
Also, it's likely that the validators' API will be modernized with my work on promises, so best to put your PR on hold.
PS: Thanks for the PR

johanwikstrom:
Sure, np! Thanks!

Issue #916, opened 2015-05-25
Bug: Error messages from parsley.remote removed when client-side validation succeeds.


pcuriel:
Hi everyone.

I have experienced issues with fields that have both client-side validation and server-side validation with parsley.remote.js. 

With this example the error can be reproduced: http://jsfiddle.net/pcuriel/816438wa/

1) Trigger validation without modifying the "Email" field. This way, client-side validation should succeed but the remote one should not.
2) Trigger validation once more and without modifying anything again. This second validation should be triggered having the focus on the "Email" field and pressing the "Enter" button to reproduce the error, because when validation is triggered pressing the "Submit" button everything works as expected.
Now the remote validation fails again, but both the error message and the error style appears briefly and then disappears.
3) Steps 1 and 2 can be repeated with a different email (all emails beginning with "a@" are considered invalid) with a similar result: in the first attempt the error message/status will correctly appear and the second will not.

I haven't had time to look more carefully into the source code for the cause of the error. But taking into account that the error is shown the first time a given value is tested, I have tried clearing the remote validation cache (check this example http://jsfiddle.net/pcuriel/o31mttdc/, where source code of parsley.remote.js is modified to do so -lines 230 to 232-) and the problem is solved.

When remote validation uses cached results, are this results "computed" before the client-side validation? If so, could it be that client-side validation clears this errors when it succeeds?

pcuriel:
Answering to my own question, the problem is not cached remote validation results being hidden by client-side ones. I have found out that when a second validation is triggered with the "Enter" button, actually two validations are run, but the second does not seem to fire the remote validation.

This can be observed in the following screenshot of the JS Console (where it says "Line X" it means that the log trace has been generated from Line "X" of parsley.remote.js).

![parsleybug](https://cloud.githubusercontent.com/assets/1854183/7813805/3496f96a-03be-11e5-8132-42b1116d4410.png)

- The traces in (1) have been generated after triggering the first validation in the form.
- The traces in (2) have been generated after triggering a second validation in the form using the "Enter" button. It can be observed that a second validation is run but without async validation being called, and thus validation result is success and the error message added in the first validation is removed.
- The traces in (3) have been generated after triggering a third validation but clicking in the "Submit" button. In this case no second validation is run and thus the error message is not removed.

[...]

PR #921, opened 2015-06-08
Issue 916 kludge

JanezStupar:
This is a PR for issue #916. It is a kludge-y approach to fixing a race condition that happens when remote validators are used on fields which also utilize local validation (remote validation fails, displays error message -> parsley binds validation to keyup event) -> which leads to situation where error message gets hidden immediately after triggered by 'enter' keyup.

The issue is resolved by surpressing keyup handler, whenever it is triggered by Enter keyup event (no need to validate on enter keypress).


marcandre:
Thanks a lot for the PR.
I'm actively working on a change on handling validations (remote and local) using promises which I hope would resolve the issue by itself. I'll keep this opened to check in case it doesn't though.

JanezStupar:
@marcandre that is great news. I feel your approach is the correct way to go.

jayepoch:
Thanks for your help on this guys! Experiencing this bug too & wondering how the fix might be coming, @marcandre? Anything we can help with?

marcandre:
>  Anything we can help with?

Yes! Haven't had time to check it out, but anyone can try out the v2.2.0rc1 release and let me know if the issue is still there or not... :smile:

jayepoch:
Thanks Marc! We tried RC1 but it unfortunately broke some aspects of our
validation process, so we're using the kludge fix for now

On Mon, Aug 17, 2015 at 4:13 PM, Marc-André Lafortune <
notifications@github.com> wrote:

> Anything we can help with?
>
> Yes! Haven't had time to check it out, but anyone can try out the
> v2.2.0rc1 release and let me know if the issue is still there or not... [image:
> :smile:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guillaumepotier/Parsley.js/pull/921#issuecomment-131989100>
> .
>



-- 
Jason Wohlstadter
CEO, Proton
sndsystm.com/+

marcandre:
> Thanks Marc! We tried RC1 but it unfortunately broke some aspects of our
validation process

It would be great if you could elaborate on that; v2.2 is meant to be compatible, so if you could provide me with a way to check it out or more details on what is breaking, I could see what I can do.

As you can guess, we won't support older versions so it's in your best interest to upgrade at some point...

tiimbz:
> It would be great if you could elaborate on that; v2.2 is meant to be compatible, so if you could provide me with a way to check it out or more details on what is breaking, I could see what I can do.

When upgrading from v2.0.7 to v2.2 the Parsley.remote validation breaks. Other validations work fine, it's just remote that seems to get skipped. When submitting the form, the form does its normal validation like parsley-required, but no parsley-remote request is fired to the server. 

This is the code:
```
<input type="text" id="proton-id-val" name="protonid" class="form-control"
       data-parsley-trigger="focusout"
       data-parsley-errors-container="#proton-id-error"
       data-parsley-remote-reverse="true"
       data-parsley-remote="http://api.domain.com/api/validate/proton_id"
       data-parsley-remote-message="Taken"
       data-parsley-required-message="Required"
       required/>
```

marcandre:
>  no parsley-remote request is fired to the server

Oh my :bomb:  I'll fix that ASAP (tonight probably).

marcandre:
So I had trouble investigating because or [#956]. This hasn't changed though, so should not be a difference.

That said, the request is sent. Here's [a fiddle that works](http://jsfiddle.net/marcandre/ke8ebpcp/), based on your code....

Be sure to have your console open before your field gets the focus, though, coz with your trigger, and *the fact that Parsley caches requests*, you might not see a request when you expect one...

marcandre:
Just to be clear @tiimbz, I'd be very grateful if you could confirm you actually see a difference between 2.1 and 2.2, and if so, tell me more about it because it is meant to be compatible.

tiimbz:
@marcandre Your fiddle indeed worked without any issues. The moment I replaced parsley.remote.js from v2.1.3 to v2.2.0 in our app, all AJAX calls broke. 

I was able to get everything working again by switching the order of which parsley.js and parsley.remote.js were loaded. Our app loads parsley.remote.js **BEFORE** loading parsley.js (as described in the documentation), but the problem disappeared by loading parsley.remote.js **AFTER** parsley.js. 

On topic: I could finally test this version against this bug #916 , and can confirm that with 2.2 indeed fixes it :) The error message is not hidden anymore on pressing Enter, and stays visible like it should, just like with clicking submit.

marcandre:
You're supposed to load either `parsley`, or `parsley.remote`, not both :smile:

We'll probably merge them very soon, so confusion will disappear.

I'm thus closing this, will reopen if there's additional issues. Thanks!

tiimbz:
> You're supposed to load either parsley, or parsley.remote, not both :smile:

Maybe good to also update the documentation on to reflect this? 

From http://parsleyjs.org/doc/index.html#remote: "To use this plugin, either load parsley.remote.js before loading parsley.js, or directly load parsley.remote.js."

marcandre:
Indeed, I tweaked the doc, thanks

PR #923, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "parsleyjs",
-  "version": "2.1.0-rc10",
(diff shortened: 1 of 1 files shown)


guillaumepotier:
Hi there, why not.

But you'll have to remove it too from [here](https://github.com/guillaumepotier/Parsley.js/blob/master/Gruntfile.js#L26) because this is automatically added while doing grunt build command.

Best

guillaumepotier:
Done in upcoming 2.1.2. Thks !

PR #926, opened 2015-06-11
Re-bind custom triggers during a reset

rphelan:
This is related to #843 which deals with the automatic change trigger but does not address the fact that when you call reset() any custom bindings added from the data-parsley-trigger attribute are lost.


guillaumepotier:
Thanks a lot for this great PR!

Issue #930, opened 2015-06-22
Parsley is undefined when loading from minified file?


Download:
I'm pretty new to AMD / require.js as well as to Parsley, so excuse me if this turns out to be a user error...

I'm loading Parsley through AMD using require.js, but for some reason I'm getting `undefined` as the Parsley object.

Here is the strange thing: If I load the source file of Parsley (I have to add some paths so it can find it's own sub modules) it does work!

I created a [JS Fiddle](http://jsfiddle.net/01cg32p8/1/) that demonstrates my issue, but for some reason I could only get this to work using `require`. In my actual code I'm using `define`

Here is my code using the minified script:
```javascript
require.config({
	paths: {
		'jquery': '//code.jquery.com/jquery-2.1.4.min',
		'parsley': '//cdn.rawgit.com/download/Parsley.js/2.1.2/dist/parsley.min',
	}
});
define(['parsley'], function(Parsley){
	alert('Parsley is ' + Parsley);   // ---> alerts: 'Parsley is undefined'
});
```

If I use the source version of Parsley, it does work:

```xml
require.config({
	paths: {
		'jquery': '//code.jquery.com/jquery-2.1.4.min',
		'parsley': '//cdn.rawgit.com/download/Parsley.js/2.1.2/src/parsley',
		'i18n': '//cdn.rawgit.com/download/Parsley.js/2.1.2/src/i18n',
		'validator': '//cdn.rawgit.com/guillaumepotier/validator.js/1.1.2/dist/validator.min'
	}
});

define(['jquery', 'parsley'], function($, Parsley){
	alert('Parsley is ' + Parsley);   // --->  alerts 'Parsley is [Object object]'
});
```

If I leave out the dependency on jQuery, like so:

```javascript
define(['parsley'], function(Parsley){
	alert('Parsley is ' + Parsley);
});
```

I get two errors:
* Uncaught ReferenceError: jQuery is not defined (anonymous function) @ en.js:6
* Uncaught ReferenceError: $ is not defined (anonymous function) @ pubsub.js:7

I think this is a bug. `en.js` and `pubsub.js` should `define` themselves and declare the dependency on jQuery...

If I add the explicit dependency on jQuery to the define using the minified Parsley, it doesn't actually help. Neither does adding the paths for `i18n` and `validator`. Am I doing something wrong?

Download:
[...]

PR #931, opened 2015-06-23
Make AMD module in minified file return Parsley 

Download:
That way it's passed to the define callback function of dependant modules as expected.
Fixes https://github.com/guillaumepotier/Parsley.js/issues/930


guillaumepotier:
Thanks @Download!

Best

Download:
Thank you!!  :+1: 

My name is Stijn BTW  :)

PR #935, opened 2015-07-09
Add filter options

frntz:
Adding a filter option allows us to select more precisely inputs we need to validate, especially when facing validation with complex components like select2 (we need to validate hidden type input which are "visible" due to dom parents visibility).
We achieve this case in Parsley1 by returning true/false in onFieldValidate which is impossible in Parsley2, adding filter option covers our need.


marcandre:
It might not be clear enough, but you can already specify the `excluded` option on each individual field.

Also, while it is not specified, you can already pass a function as the `excluded` option of a form.

Are both of these sufficient for your needs?

marcandre:
PS: third option is to create a jQuery pseudo selector and use that in `inputs` or `excluded`...

frntz:
thanks @marcandre setting a function inside the excluded option works !

PR #942, opened 2015-07-21
Fix typo fist to first for gh-pages

jonsuh:


guillaumepotier:
Hi @jonsuh thanks for your PR. Unfortunately, we never commit on `gh-pages` we just merge `master` there, so the modifications on `master` would override these ones.

Best

Issue #826, opened 2015-01-16
Submit Button Value not being submitted due to Parsley


fabianabela:
I have encountered an issue where the value of a submit button is not being submitted with the form data. This is happening because of parsley. After some more troubleshooting I have discovered that if I make use of the script parsley.js and not parsley.remote.js, the issue is not there anymore.

Note: This issue only happens in Internet Explorer and Firefox, while in Chrome everything works just fine.

Working example

Here is an example in plunkr which demonstrates that with the script - parsley.js, this issue does not happen in neither of the browsers, see: http://plnkr.co/edit/0ylxasiyrlxhSrhrMaFH?p=preview

This means that if we run the above plunkr in any browser and enter anything in the text box and then click on submit, the submit button value is going to change from "undefined" to "submitBtn" as it should.

Not working example

Here is another example in plunkr which demonstrates that with the script - parsley.remote.js, there is this issue in both Firefox and Internet Explorer, while it is fine in Chrome, see: http://plnkr.co/edit/ODqZannjYFBEwppXOXiF?p=preview

This means that if we run the above plunkr in either internet explorer or firefox and enter anything in the text box and then click on submit, the submit button value is never going to change to "submitBtn" as it should, but instead it will always stay "undefined".

Anybody has ever encountered this problem or have a solution? Thanks a lot.

marcandre:
First of, **thank you** for providing a working example of the issue!

This is a problem, and in all browsers this will also cause issues with forms with multiple submit buttons.

It's also not easy to solve, as there is no way to know what causes a submit event. We can check the `click` events of submit buttons and remember the value, but I'm not even sure what's the best way to submit the right information next.

fabianabela:
@marcandre thanks a lot for your response. To be honest, I didn't completely understood your reply. Note, this issue does not occur in Google Chrome (I am re-saying this, since you have said - "in all browsers"). Do you think there is a temporary solution that I can adopt in order to fix this please?

lekoala:
If you are ok for an ugly solution to fix this meanwhile, you can add this when you initialize parsley. The idea is indeed to register the submit handler and then submit it as a hidden element. Not pretty, but works for me.

		// Add action to submitted form for parsley.remote
		// See issue https://github.com/guillaumepotier/Parsley.js/issues/826
		var submitActor = null;
		$('form.parsley [type=submit]').click(function() {
			submitActor = $(this);
		});
		$.listen('parsley:form:success', function(formInstance) {
[...]

PR #947, opened 2015-08-03
Promise

marcandre:
This branch's goal is for the `remote` plugin to become a "normal" validator. It also removes the dependency on the `validators` library.

It achieves this by allowing validators to return promises in addition to `true` and `false`.

An improved API for custom validators is proposed, with the previous API being maintained for compatibility.

The new API makes it clear if the validator is to validate multiple values (checkboxes) or single values (or both), proposes a builtin way to parse the requirements.

As I wanted to make `remote` a normal validator, I needed to provide access to extra parameters, so `validatorKind` can be an object, with keys being the different parameters and the values being the desired type. It works well, but I realized after doing this that I still needed access to the name of the field.  So I am now passing the parsley instance as the last parameter. Had I thought of that, I might not have written the more complex "extra options" parsing, but now that it's written, it could probably be helpful to others (or is it too complex for nothing?)

One thing I'm considering is to move the `addValidator`/`addMessage` to `window.Parsley`.

I must admit that, contrary to my normal habits, the commits are not perfectly atomic, but I don't have the courage to clean that up.

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -168,12 +168,12 @@
 &lt;script src="i18n/it.js">&lt;/script>
 &lt;script src="parsley.min.js">&lt;/script>
 &lt;script type="text/javascript">
-  window.ParsleyValidator.setLocale('fr');
+  window.Parsley.setLocale('fr');
 &lt;/script>
 </code></pre>
               </li>
               <li>
-                Load your needed localization <strong>after</strong> Parsley. The last loaded file will automatically set the messages locale for ParsleyValidator. In this example, we'll load both French and Italian translations, and use Italian:
+                Load your needed localization <strong>after</strong> Parsley. The last loaded file will automatically set the messages locale for Parsley. In this example, we'll load both French and Italian translations, and use Italian:
 <pre><code>&lt;script src="jquery.js">&lt;/script>
 &lt;script src="parsley.min.js">&lt;/script>
 &lt;script src="i18n/fr.js">&lt;/script>
@@ -776,7 +776,7 @@ if (false === instance.options.priorityEnabled)
 <pre><code>&lt;input type="text" data-parsley-multipleof="3" />
 [...]
 &lt;script type="text/javascript">
-window.Parsley.Validators
+window.Parsley
   .addValidator('multipleof', {
     requirementType: 'integer',
     validateNumber: function(value, requirement) {
--- a/doc/index.html
+++ b/doc/index.html
@@ -781,10 +781,12 @@ window.Parsley
     requirementType: 'integer',
     validateNumber: function(value, requirement) {
       return 0 === value % requirement;
+    },
+    messages: {
+      en: 'This value should be a multiple of %s',
+      fr: 'Cette valeur doit être un multiple de %s'
     }
-  )
-  .addMessage('en', 'multipleof', 'This value should be a multiple of %s')
-  .addMessage('fr', 'multipleof', 'Cette valeur doit être un multiple de %s');
+  );
 &lt;/script>
 </code></pre>
           </p>
@@ -857,6 +859,10 @@ window.Parsley
 
           <p>For even cases where more complex parameters are needed, you can specify extra parameters; refer to the source and check how the <code>remote</code> validator uses that.</p>
         </div>
+
+          <h3 id="custom-messages">Error messages</h3>
+          <p>You can specify error messages, in as many locales as desired, using the <code>messages</code> option.</p>
+          <p>This is equivalent to calling <code>addMessage</code> for each locale.</p>
 
         <!-- ****************** UI/ UX ****************** -->
         <h1 id="ui" class="page-header">UI/UX</h1>
@@ -1372,6 +1378,7 @@ $('[name="q"]').parsley()
                   <li><a href="#custom-intro">Craft yours!</a></li>
                   <li><a href="#custom-function">Validating functions</a></li>
                   <li><a href="#custom-requirement">Requirement parameters</a></li>
+                  <li><a href="#custom-messages">Error messages</a></li>
                 </ul>
               </li>
               <li>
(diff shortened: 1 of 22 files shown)


doc/index.html
```diff
@@ -769,51 +767,95 @@ <h3 id="validators-list">Validators list</h3>
             These validators are shipped in <code>parsley.js</code>. Have a look at the <a href="#remote">Parsley Remote</a> plugin and <a href="#extras">Parsley Extras</a> for more validators.
           </p>
 
-          <h3 id="validators-craft">Craft yours!</h3>
+          <h1 id="custom">Custom Validators</h3>
+          <h3 id="custom-intro">Craft yours</h3>
           <p>Of course, Parsley built-in validators are commonly used validators, and you'll need some more that fit your specific forms and validations. That's why Parsley lets you easily create your own validators.</p>
           <p>
-          <p>Here again, like localizations, configuring your custom validators and error messages comes in two flavors:
```
guillaumepotier: No more support for global validator registration?

src/extra/plugin/remote.js
```diff
@@ -6,12 +6,9 @@
 // Because returns promises instead of booleans.
 (function($){
 
-window.ParsleyExtend = window.ParsleyExtend || {};
-window.ParsleyExtend = $.extend(true, window.ParsleyExtend, {
-  asyncSupport: true,
-
+$.extend(true, window.Parsley, {
```
guillaumepotier: We should remove/update comments above right?

src/parsley.js
```diff
@@ -14,7 +14,7 @@ define([
   // An abstract class shared by `ParsleyField` and `ParsleyForm`
   'parsley/abstract',
   // A proxy between Parsley and [Validator.js](http://validatorjs.org)
```
guillaumepotier: Sould be updated too ;)

src/parsley/field.js
```diff
@@ -26,24 +26,36 @@ define('parsley/field', [
...
       this._trigger('validate');
 
-      this._trigger(this.isValid(force, this.value) ? 'success' : 'error');
-
-      // Field validated event. `this.validationResult` could be altered for custom needs too
-      this._trigger('validated');
-
-      return this.validationResult;
+      return this.whenValid(force, this.value)
+      .done(function()   { that._trigger('success'); })
```
guillaumepotier: Personal taste: when I use promises I usually increment these methods to show more distincly they results to action above. Find that more readable, but again just personal taste.

src/parsley/field.js
```diff
@@ -63,37 +75,53 @@ define('parsley/field', [
...
-          return false;
-
-      return true;
+        return $.when();
+
+      var groupedConstraints = this._getGroupedConstraints();
+      var that = this;
+      var promises = [];
+
+      $.each(groupedConstraints, function(_, constraints) {
```
guillaumepotier: I think this block which is very important lacks some comments and overall readability. (I'll take a closer look once I'll read all the other modifications, just going through files from page beginning to end)

src/parsley/form.js
```diff
@@ -15,70 +15,100 @@ define('parsley/form', [
...
     },
 
-    // @returns boolean
+    // internal
+    // _submit submits the form, this time without going through the validations.
+    // Care must be taken to "fake" the actual submit button being clicked.
+    _submit: function() {
+      if (false === this._trigger('submit'))
+        return;
+      this.$element.find('.parsley_synthetic_submit_button').remove();
```
guillaumepotier: Not sure to follow here

test/features/remote.js
```diff
@@ -10,113 +10,76 @@ define('features/remote', [
 
   return function () {
     describe('ParsleyRemote', function () {
+      var stubbed = false;
```
guillaumepotier: :+1:

doc/index.html
marcandre: No more support for *local* validator, because it never actually worked, there was only a single global dictionary of validators. I think local validators are overkill, but they could be rolled in if you think it's helpfull.

src/extra/plugin/remote.js
marcandre: Indeed, done.

src/parsley.js
marcandre: :+1:

src/parsley/field.js
marcandre: :+1:

src/parsley/field.js
marcandre: :+1:
I created a private method and added comments

src/parsley/form.js
marcandre: Not too sure how to explain better, but this is to resolve #826

guillaumepotier:
> The new API makes it clear if the validator is to validate multiple values (checkboxes) or single values (or both), proposes a builtin way to parse the requirements.

This is awesome.

> As I wanted to make remote a normal validator [...]

Should we now integrate it in `validator_registry` directly? Your PR reduces code base size, we could afford to ship it by default and remove a version to build / test / and maintain? Do you want me to work on that and make a PR? 

> One thing I'm considering is to move the addValidator/addMessage to window.Parsley.

Why not, while maintaining current API and deprecate it in an upcoming version.

When you are ready, ping me, I'll make builded versions, update doc/bower/npm tag and release :)

Best

src/extra/plugin/remote.js
```diff
@@ -1,271 +1,104 @@
...
-
-  asyncValidators: {
-    'default': {
-      fn: function (xhr) {
-        return 'resolved' === xhr.state();
+(function ($) {
+
+  $.extend(true, window.Parsley, {
+    asyncValidators: {
+      default: {
```
guillaumepotier: I think leaving `default` not in a string form could really mess with IE8 as it is a [reserved word](https://mathiasbynens.be/notes/reserved-keywords).

marcandre:
Ok, this should be releasable. @guillaumepotier I've added ~5 commits, should be easy to check.
I've not done the `remote` merge, but if you're ok with this, we should release this as 2.2.0rc1

guillaumepotier:
This is good to me yes. I'll release 2.2.0-rc1 then :)

PR #951, opened 2015-08-07
Remove duplicated code in comparison.js

### LINKED ISSUE #950: Not equal comparison code is wrong
Hi @marcandre, @mvpotter and @vsn4ik,

The comparison code located in here https://github.com/guillaumepotier/Parsley.js/blob/master/src/extra/validator/comparison.js#L45-L51, is just a copy-paste of "Less than or equal to validator".

FYI. Have a good weekend.

vsn4ik:
Fix #950.

The extra code added in daa43335099eb5caf19a63025922b4f37f0c4e6d.

### DIFF
--- a/src/extra/validator/comparison.js
+++ b/src/extra/validator/comparison.js
@@ -6,7 +6,7 @@ var parseRequirement = function (requirement) {
     return +requirement;
 };
 
-// gt, gte, lt, lte, notequalto extra validators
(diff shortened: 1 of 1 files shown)


vsn4ik:
Removed in 416afc5a6426175999ee1d8a4a56aaedd6a8f0b4

PR #952, opened 2015-08-08
Promise review

guillaumepotier:
Hi @marcandre, here is a small review of your great PR. Nothing much to say, this is great stuff.

I just fixed here and there these kind of syntax I use across this project:

```javascript
var name = function(arguments) { 
   // ...
}

// use that instead
var name = function (arguments) { 
   // ...
}
```

```javascript
if(condition) {
    // ...
}

// use that instead
if (condition) {
    // ...
}
```

and new hype I'm currently on various projects is using `replace()` over concatenation for messages

```javascript
throw 'Value "' + name '" is invalid';

// find that more readable
throw 'Value %s is invalid'.replace('%s', name);
```

I'll add a comment on your PR for your questions.

Best

### DIFF
--- a/src/parsley/validator.js
+++ b/src/parsley/validator.js
@@ -22,6 +22,12 @@ define('parsley/validator', [
         throw 'No such reference: "' + string + '"';
       return result;
     },
+    boolean: function(string) {
+      return string !== 'false';
+    },
+    object: function(string) {
+      return ParsleyUtils.deserializeValue(string);
+    },
     regexp: function(regexp) {
       var flags = '';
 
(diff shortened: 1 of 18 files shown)


marcandre:
For the spaces, that's fine. For the replace, if you're ok with going with ES6, then there is string interpolation builtin...   `Value ${name} is invalid`

src/parsley/validator.js
```diff
@@ -40,79 +43,96 @@ define('parsley/validator', [
...
-    var m = string.match(/^\s*\[(.*)\]\s*$/)
-    if (!m)
-      throw 'Requirement is not an array: "' + string + '"';
-    var values = m[1].split(',').map(ParsleyUtils.trimString);
-    if (values.length !== length)
-      throw 'Requirement has ' + values.length + ' values when ' + length + ' are needed';
-    return values;
+  var convertArrayRequirement = function (string, length) {
+    try {
+      var array = $.parseJSON(string);
```
marcandre: This only passed the tests because I mistakenly used `eql` instead of `equal` in `test/validators.js` and that my example happened to not have strings, but I want to parse `[hello, world]` which is not JSON.
I'll restructure this PR so we can pick and choose better.

marcandre:
OK, I've reorganized the commits & merge the bulk of it.
I'm :-1: on the two remaining commits. The last one coz it's ugly and superceded by better in ES6.
The first one I'll comment inline, but the summary is "in those cases I prefer being explicit"

marcandre:
In short: I won't fight to death over these small commits, merge what you like :smile_cat: 

PS: I took the liberty of setting the author to you, although you didn't write the comments...

rtmalone:
Looking forward to seeing this in the code base soon (hopefully!). Working with custom validators now and the documentation vs examples on parsleyjs.com don't seem to agree with each other. My hope is that this PR by @marcandre will clear that up. Loving this project.

marcandre:
This has been merged :smile: although there seems to be a major problem with the `remote` validator I'll handle tonight...

It's very possible some examples are using the older API. If you can point out the ones you saw that would be helpful.

rtmalone:
Of course! 

The [custom validator demo](http://parsleyjs.org/doc/examples/customvalidator.html) seems to be using the old API

A small typo in the documentation [here](http://parsleyjs.org/doc/index.html#psly-custom-requirement) I believe should be `requirementType` instead of `requirementKind`

Initially, I was using 2.1 and the `.addValidator` method was not being recognized, so I checked the versions available for download and the annotated source. After updating to 2.2.0-rc1 it appears registering a new validator is working

marcandre:
Oh my, the custom validator requires more than fixing the API, actually... Opened #957.

Fixed the typo, thanks.

Otherwise the current doc should always corresponds to the latest version.

Issue #717, opened 2014-08-22
Move i18n to dist


bzmwillemsen:
I was wondering if the production build could be expanded to include the i18n files?

I'm would like to add the i18n files to jsDelivr, but they would prefer that they be minified and approved for release by being in the dist folder.

Thoughts?

cc: https://github.com/jsdelivr/jsdelivr/pull/1502

guillaumepotier:
I'll look closer your PR this week end, it sounds really good though :)

Best

marcandre:
The latest release now has `dist/i18n` :smile:

PR #964, opened 2015-08-25
provide UMD-wrapped i18n files

fastfrwrd:
For users of Parsley that load all of their dependencies, including jQuery, from RequireJS or Browserify, shims are not sufficient when loading the i18n files as dependencies in their current form because of the direct usage of jQuery (which may not be available when the code is run). This commit adds a build step which copies the i18n files from src into dist and wraps them in the UMD wrapper which is used for the main parsley file.


fastfrwrd:
note, this also addresses #717 as a side benefit.

guillaumepotier:
After review this is a great PR, merging it. The i18n files should be in the dist folder in an upcoming release. Thx

PR #965, opened 2015-08-27
Power remote

marcandre:
This PR makes the `remote` validator much more flexible

1) Resurrect `field:ajaxoptions` event [#894]

2) Allow `{value}` in URL to be replaced by the value (instead of setting the data field). This is good for RESTful urls. This format follows RFC 6570 on URI templates.

@guillaumepotier I assume you're :+1: on these?


guillaumepotier:
Yes indeed, this is great work, thx!

PR #966, opened 2015-08-27
Fix 'jQuery Undefined' error for CommonJS environments

larsonjj:
Switch out `jQuery` for `$` in order to fix `Uncaught ReferenceError: jQuery is not defined` when loading Parsley in a CommonJS Environment

### DIFF
--- a/dist/parsley.js
+++ b/dist/parsley.js
@@ -1690,7 +1690,7 @@
 window.ParsleyConfig = window.ParsleyConfig || {};
 window.ParsleyConfig.i18n = window.ParsleyConfig.i18n || {};
 // Define then the messages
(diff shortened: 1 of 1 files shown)


marcandre:
@guillaumepotier is working on restructuring a bit our source code and move it to ES6. Then intl files should be part of that too I think.

guillaumepotier:
Right. Furthermore, modifying `dist/` files is useless since it would be replaced by the next build we'll make when releasing a new version ;)

marcandre:
Let's keep this open until it's resolved, though, just in case we forget.

marcandre:
Should be all good in latest release. Thanks!

PR #998, opened 2015-10-08
Use $ instead of jQuery object while using parsley with requirejs. The jQuery object is not defined under the noConflict mode which leads to failure while accessing the extend method.

panalbish:


marcandre:
You're right, sorry about this.

I'm fixing these in the `es6` branch in a different way.

panalbish:
Thanks for that. Is there any roadmap for the new release?

marcandre:
`2.2.0-rc4` has been released...

PR #1000, opened 2015-10-13
Corrected some grammar in Norwegian translation

bjarte:
The sentences were, in my opinion, a too direct translation from English. This caused a few of the sentences to sound a bit odd.


marcandre:
THanks

Issue #1008, opened 2015-10-23
On number fields an entry of - is validating


subssn21:
When using a number validator an entry of - (negative sign) with no number entered passes validation.

marcandre:
Would be nice to address. PR welcome

subssn21:
I'll see what I can do, already worked around the issue by adding a custom validator, so it should be easy enough to add to the number validator.

subssn21:
I have a working fix and a test, but when I tried to run the test I get
this error.

I cloned the repo and ran the instructions from the site, I am sure I am
just missing something silly.

mocha was not initialized before the page finished loading. Make sure to
include mocha.js as a direct script and call `mocha.ui` or `mocha.setup`.

On Fri, Oct 23, 2015 at 10:35 AM, Marc-André Lafortune <
notifications@github.com> wrote:

> Would be nice to address. PR welcome
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guillaumepotier/Parsley.js/issues/1008#issuecomment-150591970>
> .
>



-- 
Michael Robellard
(216)288-2745

marcandre:
Sorry for the troubles you're having. `mocha` is installed when you do `npm install`. Did you get an error then?

You can create a PR anyways, the tests will run on travis. In case of errors, amend you commit and force push...

subssn21:
I think yhis commit from Aug 29, 2015 broke your Travis CI integration. No
once pull request have passed since then and they all have the same error
as the one I saw ths morning when I was trying to run the tests myself:



Commit: 2b96eb3efe0db74a88daf1a3d3e49f88e0c30a44

On Fri, Oct 23, 2015 at 11:48 AM, Marc-André Lafortune <
notifications@github.com> wrote:

> Sorry for the troubles you're having. mocha is installed when you do npm
> install. Did you get an error then?
>
> You can create a PR anyways, the tests will run on travis. In case of
> errors, amend you commit and force push...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guillaumepotier/Parsley.js/issues/1008#issuecomment-150613409>
> .
>



-- 
Michael Robellard
(216)288-2745

PR #1009, opened 2015-10-23
Fix for - only being allowed in number validator

subssn21:
Changed the number regex so that you must have at least a number with an integer part or a number with only a decimal part for a number to be valid, thus making only a number with only a - (negative sign) and invalid combination

Fixes: https://github.com/guillaumepotier/Parsley.js/issues/1008


subssn21:
I think this commit from Aug 29, 2015 broke your Travis CI integration. 

Commit: 2b96eb3efe0db74a88daf1a3d3e49f88e0c30a44

No one's pull request has passed since then and they all have the same error as the one I saw ths morning when I was trying to run the tests myself with latest pull from master.

subssn21:
Trying again in a minute

PR #1010, opened 2015-10-25
Minor change to get rid of deprecation warning

AWessel77:
The warning says that the method 'addCatalog' shall be invoked on Parsley instead of ParsleyValidator.


marcandre:
Thanks for the PR. All fixed in `master`

PR #1011, opened 2015-10-26
Adding hr locale

nhenezi:
Adding Croatian translation of validation messages


marcandre:
Thanks for the PR. The way we write i18n files have changed a bit, they're now under `/src/i18n/`, so you commit can't be merged as is. Would be great if you want to modify it, otherwise I'll do it manually

marcandre:
Merged manually. Thanks

PR #1023, opened 2015-11-16
Set french locale after adding the messages

lammensj:
When including the french version of Parsley, the messages are not being translated because the locale is not set.


marcandre:
Thanks for the PR.

Before I merge it, I need to tweak the `src/i18n` to import Parsley and remove it for `dist/i18n`, otherwise the tests fail...

marcandre:
Merged manually.
Thanks.

PR #1064, opened 2016-02-16
Update es.js

caspii:
A spanish user claims that: 

"Este valor es requerido" which sounds not natural into spanish.

Please replace by:
"Tienes que informarlo, es obligatorio."


marcandre:
Muchas gracias para este pull request. The thing is, all the other error messages have the same form, which also avoid the issue of having to choose between "Tiene" and "Tienes".
So I think it's best to keep the error message as is, but of course you're welcome to change it for you own use.

caspii:
OK, sure

Issue #1067, opened 2016-02-22
remote validation breaks when `data` is `FormData()`


papiro:
I've got an `$.ajax` options object which looks as such:

    {
            contentType: false,
            headers: {
                "Authorization": "Bearer " + WAMSession.get("token")
            },
            method: "PUT",
            processData: false,
            url: ENVCONFIG.serviceRoot + "/images?" + $.param(serialData),
            data: imgFormData
    }

`imgFormData` is a `new FormData()` instance with an image binary appended.

What happens is it gets run through `$.param` (in remote.js source) when trying to build a key for the remote caching.

I alleviated the issue locally by removing the 5 or so lines of code which implement the remote caching, and I wondered, what was the original intent of the remote caching?  Is it a fix to a bug or a feature?

If it's a feature, I thought about making a pull request which allows for a parameter like `{ remoteCache: false }` or something like that.

Thanks,
-Pierre

marcandre:
The caching is definitely a feature, to avoid hits on the server for values that have already been tried. Ideally `parsley` would be `FormData` aware and build the cache key from the form data. Would that work for you?

papiro:
That would work, for sure.  Just not sure how I would implement that.  Any ideas?  Also, what if the UI wanted to try the same exact call more than once?  Like what if the validation depended on some variable on the server?  Does that sound weird?

marcandre:
Yeah, it does sound weird :smile:  We clear the cache after a successful validation, btw.
So for example if there's a check for a duplicate email on account creation, then that will still work.
So it's basic cache management, but I've yet to see a valid example where it doesn't do the job...

Haven't tried it, but if `$.param` chokes on `FormData`, you could find a way to serialize it yourself and then call `$.param` on a copy of `ajaxOptions` where the `data` field has your serialized version.

Are you feeling like creating a PR for this?

papiro:
Well, I definitely don't mind doing a PR for the fix.  I'm just wondering - in my case I'm appending binary data.  Can that be serialized?  Should it be base64-encoded before serialization?  How about being able to "turn off" remote caching on a case-by-case basis?

marcandre:
Wow, sending binary for validations? That's using the FileAPI, right?
Just curious: are you validating whole files?
For binary files best thing would be to serialize using a hash (a simple one, like [this one](http://werxltd.com/wp/2010/05/13/javascript-implementation-of-javas-string-hashcode-method/))?

papiro:
[...]

PR #1074, opened 2016-02-29
Fix for sending of `FormData` by falling back to not using the Parsley remote cache

papiro:


papiro:
As part of this pull-request, an update to the documentation on the `field.validate` method was made.  It doesn't return false, but rather returns an array of validation errors.

papiro:
It's difficult for me to overcome the failure in the Travis build because just to try fixes, I have to make commits and push.  If I run the same commands Travis is running locally, I get a different result.  I'm stuck right now, without removing the reference to `FormData` from the test.

ankurk91:
I would love if this PR can be merge.

marcandre:
Mmm, sorry never commenting on this PR.

There are a couple of things going on, none of which seem clear to me.
1) useCache: what is the usecase?
2) Not too sure what's going on with `FormData`
3) Change of API for `validate`

Moreover the PR also changes `dist/`.

Because of this, I will close this and ask @papiro, @ankurk91 or anyone else to open a separate issue/PR for any of these changes that they would like to see (and why). Thanks.

papiro:
It's one year later and i had trouble remembering the issue i was having, but reading the commit messages made it clear to me.  Thankfully i wrote detailed commit messages!

marcandre:
@papiro Part of the conversation in #1067. Also see [this custom validator](http://parsleyjs.org/doc/examples/customvalidator.html) that deals with `FormData` client-side.

PR #1075, opened 2016-03-01
Allow remoteOptions to be set from a function

mrengert:
When using asyncValidator, I found it very useful to be able to load remoteOptions dynamically - in the case where the values of other form inputs are needed to validate the target input. For example, a password validator that ensures the user's personal info (name, email, etc.) is not contained in the password.

### DIFF
--- a/src/parsley/remote.js
+++ b/src/parsley/remote.js
@@ -62,7 +62,13 @@ Parsley.addValidator('remote', {
     }
 
     // Merge options passed in from the function with the ones in the attribute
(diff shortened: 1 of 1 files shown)


marcandre:
There's already a mechanism to do this, using the `field:ajaxoptions` event (check http://parsleyjs.org/doc/index.html#remote-events).
Is that enough?

mrengert:
Yes! Sorry I guess I did not expect to use events in that way. I'm used to seeing it done with callbacks. Thanks!

marcandre:
Yeah, a function was also my first thought. A long discussion in #894 later and an event was used instead. Makes it easier to compound transformations if need be.

PR #1079, opened 2016-03-10
Removing .babelrc file since Babel 6 no longer allow "blacklist" option

lgiraudel:
"blacklist" is deprecated in Babel 6.

Source: http://babeljs.io/blog/2015/10/29/6.0.0#opt-in-plugins

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "blacklist": ["useStrict"]
-}


jgentes:
UglifyJS fails to compile with this error.. 

`Module build failed: ReferenceError: [BABEL] C:\Dev\boss\boss\node_modules\parsleyjs\dist\parsley.js: Using removed Babel 5 option: C:\Dev\boss\boss\node_modules\parsleyjs\.babelrc.blacklist - Put the specific transforms you want in the `plugins` option`

marcandre:
Thanks!

marcandre:
This broke the release, see #1159.
Reverted.

marcandre:
PS: A better fix that doesn't cause similar issues welcome.

lgiraudel:
My commit is too old, I can't even remember what was my issue at the time :)

Maybe we can try with an empty .babelrc file, instead of removing it. But I think an empty file is not a valid JSON file...

gempi:
Same error here:
``
ERROR in ./~/parsleyjs/dist/parsley.js
Module build failed: ReferenceError: [BABEL] /Users/user/Projects/project/node_modules/parsleyjs/dist/parsley.js: Using removed Babel 5 option: /Users/user/Projects/project/node_modules/parsleyjs/.babelrc.blacklist - Put the specific transforms you want in the `plugins` option
``
Think you should use the babel strict mode plugin for babel 6 support (http://babeljs.io/docs/plugins/transform-strict-mode/).

Issue #1095, opened 2016-04-19
2 tests fail


joeynimu:
Hi,

When i ran the tests i get two tests failing as below;

```
122 passing (716ms)
  2 failing

  1) ParsleyForm should send submit button values, even for async validations:

      Error: expected [] to sort of equal [ 'other' ]
      + expected - actual

      -[]
      +[
      +  "other"
      +]

      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.eql (node_modules/expect.js/index.js:230:10)
      at Context.<anonymous> (test/unit/form.js:131:25)

  2)  "after each" hook for "should handle validators returning promises":
     Error: expected 1 to equal 0
      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
      at Assertion.(anonymous function) [as be] (node_modules/expect.js/index.js:69:24)
      at Context.<anonymous> (test/setup/expect_warning.js:7:46)



[11:37:14] 'test' errored after 2.87 s
[11:37:14] Error in plugin 'gulp-mocha'
Message:
    2 tests failed.
```

marcandre:
This particular test fails on Travis too. It passes on my machine, as well as in Chrome, Firefox, Safari.

I don't know why it fails on Travis and on your setup (which is?). If you do find out, let me know!

You can set the env variable 'Travis' to anything and that test will not run.

Unless someone comes up with an explanation, I will close this.

guillaumepotier:
This could be quite problematic since users submitting PR here cannot be sure what they are submitting is failing because of Travis/env config or because of their code. Furthermore, this is more reassuring to have tests passing on Travis and viewable by everyone rather than just on one specific developper laptop.

I think this open should be re-opened and ask for help on that one with Travis experts.

marcandre:
No problem, let's keep it open. I am just not investigating it. Might not be a travis issue per say, but some strange interraction of `simple-jsdom`, `jquery` and/or `node`.
@guillaumepotier do you know someone who's up for that challenge?

marcandre:
Happy to report that a clean install gives the error for me too so I can now try to understand what's going on.

benjaminpick:
The tests do not update locally (e.g. if I add failing tests), maybe this is because gulp says `gulp-notify: [Gulp notification] JSCS failed` ?

marcandre:
[...]

PR #1101, opened 2016-04-28
data-parsley-class-handler can now contain a global function name

benjaminpick:
Currently, the class handler function can only be specified as JS code:

``` js
$('#element').psly({
      classHandler: function (field) {
        return $('#field3');
      }
});
```

With this minor addition, global functions can be named in the data-attribute:

``` js
function parsleyClassHandler(field) {
        return $('#field3');
}
```

``` html
<input type="text" data-parsley-class-handler="parsleyClassHandler" required />
```

(Sorry I feel quite at loss using gulp etc. I probably need to execute the build somehow and let the test suite run, could you please add this to your CONTRIBUTING.md?)

### DIFF
--- a/doc/examples/class-handler.html
+++ b/doc/examples/class-handler.html
@@ -1,0 +1,192 @@
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="Parsley, the ultimate frontend javascript form validation library">
+    <meta name="author" content="Guillaume Potier">
+
+    <title>Parsley - Examples | Simple form demo</title>
+
+    <!-- Bootstrap core CSS -->
+    <link href="../../bower_components/bootstrap/dist/css/bootstrap.css" rel="stylesheet">
+
+    <!-- Custom styles for this template -->
+    <link href="../assets/docs.css" rel="stylesheet">
+
+    <!-- HTML5 shim and Respond.js IE8 support of HTML5 elements and media queries -->
+    <!--[if lt IE 9]>
+      <script src="https://oss.maxcdn.com/libs/html5shiv/3.7.0/html5shiv.js"></script>
+      <script src="https://oss.maxcdn.com/libs/respond.js/1.3.0/respond.min.js"></script>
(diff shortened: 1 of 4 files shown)


benjaminpick:
No idea why the travis build errors. Locally only the tests mentioned in #1095 fail

marcandre:
Thanks a lot for the PR.

I've improved the CONTRIBUTING file, I hope this clears things up.

The PR is a good start, but there are a couple of things that need to be fixed. Are you're willing to address them?
- ideally, that feature should work for any option that accepts a function callback, not just `classHandler`.
- best not to remove the functionality of accepting a direct jQuery set (even though I think it's untested and undocumented). See http://stackoverflow.com/questions/1853223/check-if-object-is-a-jquery-object
- this should probably be all one commit (squash them).
- doc needs updating
- not sure if it wouldn't be best to make separate tests instead of modifying existing ones.

benjaminpick:
* This only concerns `data-parsley-errors-container` right? `data-parsley-value` would need a new data name to avoid ambiguous situations.
* I don't think specifying the jQuery object worked before (`this.options.classHandler.call` would error). 
* AFAIK you now can squash the commits automatically when you merge the pull request (see https://github.com/blog/2141-squash-your-commits).

marcandre:
* Yeah, probably there's only `errorsContainer` too.

* The `return` was before any `this.options.classHandler.call` I think...

* ok

benjaminpick:
I have started to update the docu but I think it is best to let you finish this, you know better which parts of it need editing ... 

(You need to update the link to Contributing.md in the Readme.md)

benjaminpick:
I am getting concerns that this is also a kind of of `eval` and thus potentially enabling XSS. 

For example, the Typo3 function `GeneralUtility::removeXSS` only destroys a blacklist of HTML tags & attributes, this attribute would be left as-is. Now imagine the site has a JS function called `permanentlyDeleteAccount` that doesn't take a parameter, and some other minor flaws somewhere, and together it would be a viable XSS attack.

What do you think?

marcandre:
Oh oh, I dropped the ball on this one, sorry. Finally merged.
Thanks again for the good work!

marcandre:
BTW, @benjaminpick contact me if you'd be interested in commit access to Parsley.

benjaminpick:
I feel honored, but I will politely decline ... because I have the feeling that the code is so complex I wouldn't commit anything without discussing it with you.

PR #1103, opened 2016-05-09
Add field-level validation debouncing

molecularbear:
For remotes and other expensive validations, I wanted to be able to [debounce](http://underscorejs.org/#debounce) the validation call during user input. This works by using a debounced version of `ParsleyField.validate()` when the `data-parsley-debounce` option is specified.

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -433,6 +433,10 @@ if (false === instance.options.priorityEnabled)
                 <td><code>data-parsley-excluded</code> <version>#2.1</version></td>
                 <td>If set to <code>true</code>, Parsley will ignore this field when binding a form.</td>
               </tr>
+              <tr>
+                <td><code>data-parsley-debounce</code> <version>#2.3</version></td>
+                <td>Postpones validation for a given number of milliseconds after user input has stopped arriving, eg: <code>data-parsley-debounce="500"</code>. Useful for expensive validations (such as remotes) that you don't want to run on every keystroke.</td>
+              </tr>
             </tbody>
           </table>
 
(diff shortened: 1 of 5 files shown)


marcandre:
Interesting idea. I'm travelling right now so can't look at it, but will start thinking about it.

frg:
I've tested the pull request quite a bit and everything seems fine! :) Thanks, @molecularbear ! This is very critical to my use case since I'm being charged a small amount on a particular validation call. This is also a good temporary patch for #1120.

MicahBrown:
+1. I think remote validations should be debounced by default.

frg:
Any thoughts on this @marcandre ?

marcandre:
I took the time to look at this more carefully :smile:

So the feature is accepted and I pushed some commits I did to implement it. Released 2.5.0

@molecularbear Thanks so much for the PR. Unfortunately, there were a couple of issues with it. Main issue was that Parsley allows attributes/options to change anytime and adapts dynamically, so constructing a function at init time according to options would defeat that goal. Underscore's implementation of `debounce` is really complex (to account for `immediate=true`), it's not needed here. Also Parsley uses ES6, so no more `that = this`. Finally, as a principle, tests should not be broken in two `it`, or have side effects in general. Anyways, check out my implementation and let me know if I've missed something :smile:

@MicahBrown Maybe. Currently the `debounce` option affects any validation, not just remote ones and it's not possible in general to know which custom validations are asynch, so I'm leaving the default untouched.

frg:
@marcandre Thanks for the update! :) The time spent reviewing this is much appreciated!

molecularbear:
@marcandre I know it has been awhile, but wanted to thank you for your feedback and for working this feature into Parsley. Your changes made a lot of sense - the way you implemented the test was very clever!

PR #1121, opened 2016-06-29
Adding sourcemap to bower.json

dahliacreative:


dahliacreative:
Fixed the issue of the missing source map using bower-installers overrides options.

PR #1126, opened 2016-07-20
Defined dependencies for jspm

frg:
JSPM needs the dependencies defined in a particular format. If this is not done one would need to define the dependencies manually. Also by not defining a jQuery version jspm will use the version that is currently being used in the said project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,12 @@
       "expectWarning"
     ]
   },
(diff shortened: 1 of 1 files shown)


marcandre:
I'm a bit late to the party, sorry, but in case some people haven't switched to `yarn`...
Thanks for the PR

Issue #814, opened 2015-01-02
Date type max attribute collision with max validaton


khelll:
I want to have the attribute 'max' on date type input, and I just want to make sure parsley is validating the presence of value. However, due to the fact that parsely is using the keyword max to validate integer values, it's suggesting that the date type file is invalid although such type of validation is not required.

khelll:
Moreover, if you have numeric value in data-parsley-value, you get 'You must validate an object or a string'

MichaelAllenMiller:
:+1:

MichaelAllenMiller:
@guillaumepotier This seems like a pretty big oversight. Do you know if anyone is working on it already? If not, is there any specific way you'd like to see this implemented? Or if the preferred approach to use custom validators for date ranges?

marcandre:
Nobody is working on it, and it should definitely be fixed! The `max`, `min` and `range` validators should 
PR welcome if you feel like it:
- ideally work off the branch `es6` that I hope to merge asap.
- first idea that comes to mind would be to add a `validateDate` to the API, and add such a method to the `min`, `max` and `range` validators. Their `requirementType` will have to be string though and parsed within `validateDate` and `validateNumber`. Alternative would be to instead convert them to use a generic `validateString`?

Flaburgan:
+1 on this. It looks like there is no date validator in parsley, so actually, I guess `input type="date"` should be completely ignored to avoid collision with native validation of `min` and `max` attributes.

unilynx:
I'm running into the same problem, and am working on a fix here: https://github.com/WebHare/Parsley.js/tree/fix-814

I'm not sure how to deal with the error reporting. 

It will now return 'This value seems to be invalid.'. Ideally I would use the standard min,max and range' errors, but then I'd run into locale issues on properly formatting the date in the response.

I'll submit it as a PR anyway, as input type=date with min/max/range is currently broken anyway.

cedricnicolas:
Still not solved ? I have the issue while putting a min on input type="date"...
Uncaught Requirement is not a number: "2016-09-08"

marcandre:
I'm afraid not. There's a PR in the works, although I suspect there is a better approach to improve the range/min/max validators.

Stadly:
Any news on this?

Thumbleweed:
i've created a pull request (https://github.com/guillaumepotier/Parsley.js/pull/1158) for this one.
[...]

PR #1130, opened 2016-08-07
Adds isodatetype for input type=date support

unilynx:
Parsleyjs fails when trying to use a html5 input type=date with min and/or max. This pull request adds support for such types by adding an 'isodate' type and implementing min, max and range constraints.

I avoided the already present extra 'dateiso' type because I didn't want to conflict with something that already exists, and I think this should be working natively without the user needing to add any components or custom validators.

I think this should also fix issue #814 

The error reported will now be the defaultMessage (This value seems to be invalid.)'. Ideally I would use the standard min,max and range errors, but then I'd run into locale issues on properly formatting the date in the response. So I'm not sure how to deal with that yet, but as 'input type=date min/max=...' is currently completely broken anyway, this is still an improvement.

### DIFF
--- a/src/parsley/field.js
+++ b/src/parsley/field.js
@@ -271,15 +271,15 @@ ParsleyField.prototype = {
 
     // range
     if ('undefined' !== typeof this.$element.attr('min') && 'undefined' !== typeof this.$element.attr('max'))
-      this.addConstraint('range', [this.$element.attr('min'), this.$element.attr('max')], undefined, true);
+      this.addConstraint(this.$element.attr('type') == 'date' ? 'isodaterange' : 'range', [this.$element.attr('min'), this.$element.attr('max')], undefined, true);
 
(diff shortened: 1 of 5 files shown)


marcandre:
Awesome you're taking a stab at it. Haven't had time to look at this yet.

First question that comes to mind: did you consider/try modifying the range/min/max validators to handle dates instead of creating new validators?

unilynx:
No - I didn't see a way to easily determine whether I was dealing with an input type=date or other inside the validators, and I wanted to avoid 'if the value looks like a date, interpret it as a date'.

This seemed the easiest way to solve the issue for my use cases :-)

marcandre:
I took a stab, without resorting to a different type or different `min`, `max`, ...

If interested, have a look at #1168

marcandre:
Fixed in 2.7.0.
Thanks!

PR #1139, opened 2016-10-24
Update ja message

hiro-hori:
I felt something wrong in Japanese error message.

* This file includes misleading messages.
* Especially, `digit` message `"正しい桁数で入力してください。"` means `Please enter the correct number of digits`. It is wrong.


marcandre:
Thanks!

PR #1153, opened 2016-12-14
Improves ISO8601 date format validation

kevindb:
In addition to checking numeric _format_ of a date, this regex also checks _validity_ of dates, including leap-year (Proleptic Gregorian), works for years 1000-9999.

https://stackoverflow.com/a/28022901/1480959
Thanks to funkwurm on Stack Overflow
https://stackoverflow.com/users/2684660/funkwurm


marcandre:
Interesting. Would you care to add a simple test that this works for that the old one didn't, just for completion's sake?

marcandre:
The `dateiso` extra validator will be made obsolete when #1168 lands.

marcandre:
Fixed in 2.7.0.
Thanks!

PR #1158, opened 2017-01-02
Added 'type' attribute check on range/min/max validator

Thumbleweed:
Problem
--------
_Since the 'range', 'min' and 'max' attributes have a different meaning when using them in a date html field, parsley will fail on processing these._

This is because Parsley expects a integer, where a date field expects a string.

Solution
--------
The min/max/range attributes have no validation purposes when used on a date field.They are only used for user experience (modifying the calender popup).

My solution exists of ignoring the min/max/range validator, when attributes are present on a date field.


marcandre:
Thanks!

Issue #1104, opened 2016-05-12
Browserify and Parsley


gor0n:
Hi,

I have an issue with Parsley when I'm adding it via Browserify.

`Uncaught TypeError: $(...).parsley is not a function`

This is my script

`
var $ = require('jquery');
require('parsleyjs');

$("[data-parsley-validate]").parsley({
    trigger:      'change',
    errorClass: "parsley-error",
    classHandler: function (el) {
        return el.$element.closest('.form-item, .form-item--inline, .form-item--icon');
    },
    errorsContainer: function(pEle) {
        var $err = pEle.$element.closest('.form-item, .form-item--inline, .form-item--icon').append();
        return $err;
    }
});
`

I tried to use the window object but I got the same error.

I use Gulp with Browserify, others plugins works well with it. But I got the same error with slick carousel, then I used [the build they made especially for Browserify](https://www.npmjs.com/package/slick-carousel-browserify)  and it worked.

Thanks!

marcandre:
Mmm, I think it should be compatible but not too sure how it works. I'm travelling right now, won't have time to look it up for a while...

rhk4360:
Did you ever have any luck with this gor0n?

gor0n:
Yeah, with Browserify you can require with a path (useful to require your custom plugins that are not on npm registry).

So I just require parsley like that `require('../../../../node_modules/parsleyjs/dist/parsley.min.js')` and it works.

marcandre:
If anyone has a better fix, let me know.

rhk4360:
Unfortunately, I still get the error with that workaround so as a temporary workaround I'm loading parsley via a cdn outside of my main JS package for now. 

If I find time, I'll try to dig into the root problem a bit more. I thought it might be a problem with jquery loading after parsley because if I reference Parsley in console it's there, but it's just not tied to jquery DOM elements.

This really is a great front-end validation library so a huge thank you to the contributors.

Cheers

gor0n:
If it can help you, that's how I use parsley.

`( function( $ ) {

    "use strict";

    $("[data-parsley-validate]").length && $("[data-parsley-validate]").parsley({
        trigger:      'change',
        errorClass: "parsley-error",
        classHandler: function (el) {
            return el.$element.closest('.form-item');
        },
        errorsContainer: function(pEle) {
            var $err = pEle.$element.closest('.form-item').append();
            return $err;
        }
    });

[...]

PR #1167, opened 2017-01-30
Make jQuery a devDependency

nrambeck:
When Parsley is executed in a node environment (such as with bundlers like Browserify or test runners like Mocha), Parsley will not be properly attached to the primary instance of jQuery used in the project. 

The reason is that Parsley includes jQuery 1.11.x as a dependency, so it's internal reference to `require('jquery')` references that version. If the global version of jQuery used by the project is the same version, then there is not problem. But if a different version is used, then Parsley attaches itself to a jQuery object that is not used anywhere else in the project.

Adding jQuery to this project should only be required to run tests in a dev environment. I don't see any need to have jQuery as a production dependency. If I'm wrong then let me know and we can find another solution.

The issue this PR fixes is reported here: #1104

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,6 @@
     "url": "https://github.com/guillaumepotier/Parsley.js/issues"
   },
   "homepage": "https://github.com/guillaumepotier/Parsley.js",
-  "dependencies": {
-    "jquery": ">=1.8.0"
-  },
   "devDependencies": {
     "babel-core": "^5.2.17",
     "babel-eslint": "^4.0.5",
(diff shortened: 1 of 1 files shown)


nrambeck:
Thanks for the response. I think I just got ahead of myself and didn't dive deep enough. This is a lesson for me in how npm resolves versions for dependencies of the same package. I somehow got into a situation where Parsley was using it's own version of jQuery separate from the rest of the project, but that doesn't appear to be the fault of Parsley. Still trying to figure it out.

PR #1169, opened 2017-02-03
Add section to help with remote validation with groups

lauramosher:
After spending a few hours trying to figure out why I could not move forward on my paged/stepped form despite the form step appearing to be completely valid, I discovered it was because I added a new `data-parsley-remote` validation that was returning a promise instead of a boolean. This was causing `validate({group: "group"})` to return `null` due to a "pending validation" per your documentation, despite a `.parsley-success` in the markup.

Since the result was pretty straightforward, I thought I'd add a snippet to the `Parsley Remote` documentation section to mention to use `whenValidate()` instead of `validate()` when working with remote validations. 

Here is what it looks like once rendered (top section is what exists, bottom is new):
screen shot 2017-02-03 at 12 57 57 pm

Happy to update the text however you see fit, but I can see this being super useful to have directly in the documentation instead of all over stack overflow!

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -1202,8 +1202,18 @@ window.Parsley.addAsyncValidator('mycustom', function (xhr) {
   }, 'http://mycustomapiurl.ext');
 &lt;/script>
 </code></pre>
-            </li>
-          </ul>
+        </p>
+
+        <h3 id="remote-validation">Combining Remote Validations with Groups</h3>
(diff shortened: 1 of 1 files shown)


marcandre:
Sorry for your troubles.
Thanks for the PR, I haven't had the time to review it.
I think adding a warning in the console in case of a `null` result because of unresolved promises would be a good idea too.

marcandre:
Sorry for the delay. Merged.
I also modified the multi step examples, so we no longer have examples using `validate`.
Thanks again for the PR!

Issue #562, opened 2014-03-18
"You must bind Parsley on an existing element." on a blank page ?


rizkytahara:
Hi,

I'm upgrading parsley to version 2.

I have a blank page that load a jQuery and parsley. After that load, I got a warning in console: "You must bind Parsley on an existing element." 
Why I'm getting that warning?

Here is a fiddle: http://jsfiddle.net/xtSK5/

bnomei:
@peef is right. still there when using js init.

marcandre:
This will happen when calling `parsley()` on an empty jQuery set.
What should it do in that circumstance?

vphantom:
This error message is hard-coded at two distinct places:

- `src/parsley/main.js` line 50.  I don't see the relevance in emitting a warning when operating on a valid jQuery set that happens to be empty, which is a very common situation.

- `src/parsley/factory.js` line 30 is internally involed by the above.  Clearly if _that_ point is reached, issuing a warning does make sense.

I'll try a PR.

PR #1173, opened 2017-02-09
Don't warn on a normal but empty jQuery set

vphantom:
It is the norm for jQuery plugins to be invoked with a selector which may not have any matches on the current document.  In the case of Parsley for example, we get this warning on all pages that don't include any form inputs, which is unnecessary.

(This is in reference to #562 which addressed the DOM half.)

### DIFF
--- a/src/parsley/main.js
+++ b/src/parsley/main.js
@@ -47,8 +47,6 @@ $.fn.parsley = $.fn.psly = function (options) {
 
   // Return undefined if applied to non existing DOM element
   if (!$(this).length) {
-    ParsleyUtils.warn('You must bind Parsley on an existing element.');
-
     return;
   }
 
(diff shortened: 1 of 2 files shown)


marcandre:
I understand your intent.
May I ask some questions as to your usage though? You typically bind all forms on a page (0, 1 or many), and you prefer using Javascript than auto binding with `data-parsley`?

vphantom:
That is correct.  I use the `$('form').parsley({...})` init for consistency vs the rest of the various other components in my project, and Parsley is the only one generating such a warning on an empty set, which thus seems like a departure from a de-facto standard.

As a side-note, in my initial experiments `$.fn.parsley.defaults({...})` was ignored which may have contributed to my use of this way to instantiate it.  I haven't had time to investigate that yet though.

marcandre:
Sorry for the delay. Merged!

PR #1176, opened 2017-02-23
Fixes error if classHandler is a string but $(classHandler) has length = 0

maechler:
It may happen that the classHandler is a string but the jQuery selector has length 0 because no elements are found.
That means we do not return and later try to use "call" on the string which results in a type error:
`Uncaught TypeError: classHandler.call is not a function`

I added a type check to fix that issue.

### DIFF
--- a/src/parsley/ui.js
+++ b/src/parsley/ui.js
@@ -257,11 +257,17 @@ ParsleyUI.Field = {
   // Determine which element will have `parsley-error` and `parsley-success` classes
   _manageClassHandler: function () {
     // An element selector could be passed through DOM with `data-parsley-class-handler=#foo`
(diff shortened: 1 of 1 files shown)


marcandre:
Thanks for the PR. I agree the current behavior is bad. I'm not sure what the best behavior should be, though... It probably should be considered as a setup error and at least warn in the console... Should it return the input holder or an empty set, I'm not sure

maechler:
You are right, I will add a warning.

Hmm, I have not thought of that. I think it should probably return an empty set. Because when you set it to an element that does not exist, I would expect that no element will get the classes.

maechler:
I have updated the pull request, what do you think?

marcandre:
Oops, forgot about this one. Sorry! :blush:

Merged. Thanks again for the PR!

PR #1186, opened 2017-03-28
Remove jquery dependency from constraint.js

herberthenrique:
## Description
The file `constants.js` has dependency on jquery only for using method `$.extend`, I change it to use vanilla method `Object.assign` instead.
I also took the time to apply `const` instead of `var`.

[Object.assign - MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "blacklist": ["useStrict"]
+  "blacklist": ["useStrict"],
+  "plugins": ["object-assign"]
 }
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   "devDependencies": {
     "babel-core": "^5.2.17",
     "babel-eslint": "^4.0.5",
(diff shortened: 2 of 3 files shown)


marcandre:
Cool.

Except that `Object.assign` is not supported by IE (even IE11).
The goal is to not require any polyfill for IE9+...

marcandre:
Mmm, the simplest solution is probably to ask Babel to take care of it by adding the right plugin: https://babeljs.io/docs/plugins/transform-object-assign/

herberthenrique:
All right!  
I'll do it :)

herberthenrique:
@marcandre Sorry for the long time to code this hahaha.

marcandre:
Cool, thanks.

I replace the remaining (shallow) uses of `$.extend`

PR #1214, opened 2017-07-13
fix date field not being recognized by firefox

breizoreol:


marcandre:
Thanks for the PR!

I'm surprised this makes a noticeable difference. Does it fix the issue in IE and Firefox??

BTW, best remove the second commit; I'll take care of the release

marcandre:
(I'm travelling, so no time for testing on my end)

breizoreol:
Yes, I removed the build commit (I read the contributing documentation a tad too late :)  )

Yes it makes a difference, because "this.element.type" returns "text" on date fields in firefox... (firefox doesn't know what is a date field)

I also removed 2 mocha tests that didn't pass due to different browser behaviors (they already didn't pass before my intervention).

I tested this on firefox on a mac, not in internet explorer, but IE doesn't recognize the date type either, so it must have the same problem...

marcandre:
I went ahead and changed all uses of `.type` and built an internal function so we always get the right type and not `undefined`, since `<input>` is valid markup.
Thanks for resolving the issue!

breizoreol:
very nice !

Issue #840, opened 2015-01-29
Error messages refactor


marcandre:
There might be a good case for a refactor of the way we deal with error messages.

Potential improvements I can see:
[ ] Allow message dependent on constraint for other validators than `type` [#669]
[ ] Allow data attributes for those (e.g `data-parsley-type-number-message`)
[ ] Allow data localized attributes (e.g. `data-parsley-required-message-fr`)
[ ] Allow additional data fields (e.g. field name) [#753] [#594]

guillaumepotier:
> Allow message dependent on constraint for other validators than type [#669]

I don't really see how we could achieve something like that. using some convention to chose between a message and another for some specific validators ? Won't it add **much** complexity for only 1 or 2 tricky validators like the `length` one?

> Allow data localized attributes (e.g. data-parsley-required-message-fr) [#718]

Internationalization of custom messages is a great feature yet a tricky one. Maybe allowing users to bind an i18n function to their custom messages could be a thing ? There are plenty good javascript i18n libs out there (jed, i18next, [gettext.js :p](https://github.com/guillaumepotier/gettext.js)) and it could be good to find a way to interface with them

> Allow additional info fields (e.g. field name) [#753] [#594]

Indeed. Should not be much complex. We could use a convention in messages to allow users / some translations to replace `%value%` dynamically by the field value. Maybe some quick win here

> Allow dynamic error messages (at least for ajax) [#560]

Yes would be awesome. Should not be too complex too.

Overall, this is a good topic. Parsley errors and message handling should really benefit some lift up and improvements :)

PR #1218, opened 2017-08-18
An option to skip validation of a field if it has a certain attribute and a certain field's value

andreyyudin:
This is useful in case if you accept certain field values that would fail a normal validation.

Example: your form has "Organization" field where users put their affiliation. However, some users might have no such information. To make sure that they actually read through the form, you require them to put 'N/A' into the field. If the field's minimum length is 10, then just 'N/A' would fail validation. With this option, it would not.

### DIFF
--- a/src/parsley/defaults.js
+++ b/src/parsley/defaults.js
@@ -15,6 +15,9 @@ var Defaults = {
   // Excluded inputs by default
   excluded: 'input[type=button], input[type=submit], input[type=reset], input[type=hidden]',
 
+  // The field is excluded from validation, if it has at least one of the specified combinations of an attribute and field's value
+  excludeIfAttributeAndValue: {},
+
   // Stop validating field on highest priority failing constraint
   priorityEnabled: true,
 
--- a/src/parsley/field.js
+++ b/src/parsley/field.js
@@ -76,13 +76,26 @@ Field.prototype = {
 
   // An empty optional field does not need validation
   needsValidation: function (value) {
+    var skipValidation = false;
+
     if ('undefined' === typeof value)
       value = this.getValue();
 
+    // Determine if the field is to be excluded from validation based on its attribute and value
+    if ('undefined' !== typeof this.options.excludeIfAttributeAndValue) {
+      var thisElement = this.$element;
+      $.each( this.options.excludeIfAttributeAndValue, function( key, val ) {
+        var attr = thisElement.attr(key);
+        if (typeof undefined !== typeof attr && false !== attr && val === value)
+          skipValidation = true;
+      });
+    }
+
     // If a field is empty and not required, it is valid
     // Except if `data-parsley-validate-if-empty` explicitely added, useful for some custom validators
-    if (!value.length && !this._isRequired() && 'undefined' === typeof this.options.validateIfEmpty)
+    if ( (!value.length && !this._isRequired() && 'undefined' === typeof this.options.validateIfEmpty) || skipValidation ) {
       return false;
+    }
 
     return true;
   },
--- a/test/unit/form.js
+++ b/test/unit/form.js
@@ -183,6 +183,19 @@ describe('Form', () => {
       expect($('#element').parsley().validate(undefined, true)).to.be(false);
     });
   });
+  it('should have an excludeIfAttributeAndValue option', () => {
+    $('body').append('<form id="element" data-parsley-required>' +
+      '<input type="text" id="field1" value="baz" data-parsley-minlength="5"/></form>');
+    var form = $('#element');
+    expect(form.parsley().isValid()).to.be(false);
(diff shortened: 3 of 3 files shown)


marcandre:
Thanks for the PR.

I understand what you want to achieve, but I'm skeptical it is a way that should be encouraged (in terms of user experience). I feel that checking the value against a particular value might not be really generic enough (e.g. what about whitespace? case sensistivity?). Less importantly, I don't understand why it has to have a particular attribute to trigger the special value.

I'm not against some mechanism to bypass the validation for a particular field per say; maybe some event could be triggered from `whenValid`, but then performance might become an issue.

In short: I feel that your solution, while it probably resolves your particular need, is not generic nor useful enough for inclusion in Parsley.

andreyyudin:
Thank you for the explanation. I understand that this might be not generic enough.
The whitespace and case sensitivity are not an issue for us as we have N/A button that replaces the field value with N/A.
It has to have a particular attribute to distinguish from other fields on the page that should not accept N/A value (in case the user tries to put it in manually, there is no N/A button for those, of course).

As an alternative, please could you tell if it might be possible to have a custom validator that overrides all other validators? For example, if some of other validators fails, but this one passes, then the field value should be accepted. This should allow us to achieve the same result as with attribute/value combination.

I also have implemented a couple of other features. Please could you tell if you might be interested in any of the following (or please explain how to better use the current Parsley to achieve the same results):

 - option to skip validation if a submit button with a specific value has been pressed
 - node-specific triggers, e.g., trigger all fields' validation on focus out, but validate select fields on change
 - option to allow appending the error message instead of prepending it
 - allow changing the error message inside custom validators
 - probably the most specific of them all (and should be possible to do using the previous point): Django-like message formatting switch. Currently, fields that have min/max limitation would display just the value itself, e.g., "This value should be greater than or equal to %s.". However, Django error messages also provide the number of characters: "Ensure this value has at least %(limit_value)d character (it has %(show_value)d)". This implements the addition of show_value to the messages.

marcandre:
For you `"N/A"` issue, I presume you won't be storing "N/A" on server anyways, it's just a UI thing to make sure users don't forget to enter something when they should. So it's not actually true that you want to allow "N/A" as the input's value. You want to accept "" iff the user clicked on "N/A" and show "N/A".

The N/A button can clear the input, set the placeholder to "N/A" (or the input itself, it won't be sent because it's disabled), remove the 'required' attribute and you are done. 
That's if your N/A is a push button. But if it's a checkbox (as I think it should), then you should disable the input too.

This should achieve your goals, with the minimal difference that it doesn't allow the user to type herself "N/A" (nor should she, imo). It also reflect what is actually going on.

I'll close this PR, but conversation can continue.

> option to skip validation if a submit button with a specific value has been pressed

Well, supporting `data-parsley-novalidate` would be a good feature to add. That particular submit button should set the `novalidate` option for that form / those fields and that would work. PR welcome.

> node-specific triggers, e.g., trigger all fields' validation on focus out, but validate select fields on change
> option to allow appending the error message instead of prepending it

I believe that `errors-container` is already a versatile enough to achieve that easily, no?

> allow changing the error message inside custom validators

Already supported but not well documented. Check the [ajax example] that gives a customized error message...

OTOH, I really wish [the error messages were way more powerful](https://github.com/guillaumepotier/Parsley.js/issues/840)

andreyyudin:
Thank you for the explanation of the status of the proposed features. As 
you suggested, I was able to reproduce the editing of the error messages 
using $.Deferred() as in the ajax example.

However, I did not manage to utilize errors-container to have the error 
displayed before the field. The problem seems to be that 
errors-container always has the wrapper appended to it
if ('object' === typeof $errorsContainer && $errorsContainer.length) 
return $errorsContainer.append(this._ui.$errorsWrapper);

For example, if the field is in the table
<td>
     <input data-parsley-field-same="Passwords" 
data-parsley-maxlength="32" data-parsley-required="true" 
data-parsley-required-message="This field is required." name="password" 
data-parsley-id="15" class="parsley-error" type="password">
     <ul class="errorlist filled" id="parsley-id-15"><li 
class="parsley-required">This field is required.</li></ul>
</td>

and <td/> is being returned by errorsContainer, then the error list is 
still placed after the input. It would be necessary to have an element 
before the input and use it as the errorsContainer. I would propose a 
simple switch that would control the default behaviour to do either
return this._inputHolder().after(this._ui.$errorsWrapper);
or
return this._inputHolder().before(this._ui.$errorsWrapper);

Also, this switch could be made to influence the errorContainer.


I will prepare the data-parsley-novalidate PR.

marcandre:
Just specify an `$errorsContainer` that does a prepend and returns that.

andreyyudin:
Thank you, this works fine!

PR #1220, opened 2017-08-24
Update base.js

simkea:
Prevent trigger function for functions; make it useable for react

### DIFF
--- a/src/parsley/base.js
+++ b/src/parsley/base.js
@@ -89,7 +89,7 @@ Base.prototype = {
         if (result === false) return result;
       }
     }
-    if (this.parent) {
+    if (this.parent && typeof this.parent !== "function" ) {
       return this.parent.trigger(name, target, extraArg);
     }
     return true;


marcandre:
Thanks for the PR.
I'm confused though. `this.parent` is used by Parsley to maintain it's hierarchy from input to form to global object. The global objet's parent is `null`; in all other cases, `parent` should be an instance of `Base`. When would it ever be a function??

simkea:
in a react component!? :) i dont know.

I try to integrate your parsley in my react form.

```
import Parsley from 'parsleyjs/src/parsley';

class OrderForm extends React.Component {
    componentDidMount() {
        this.$form = $('#orderform').parsley();
    }    
    handleSubmit = (e) => {
        console.log(this.$form);
        this.$form.validate();
    };
    render() {
        const { className, orderform } = this.props;
        return (
...
                        <Form
                            id="orderform"
                        >
... some fields ...
                                    <Button
                                        type="button"
                                        variant="raised"
                                        onClick={this.handleSubmit}
                                    >
                                    </Button>
                        </Form>
...
        );
    }
}
```

if i push the submit button i get follow error log:
i have placed a console.log(this.parent); in line 93 on base.js

```
Form {__class__: "Form", element: form#orderform.mui-form, $element: jQuery.fn.init(1), domOptions: Object, options: Object…}
Base {__id__: "84", element: document, $element: jQuery.fn.init(1), actualizeOptions: null, _resetOptions: null…}
**... many rows if this elements, but then this.**..
function ( until, selector ) {
		var matched = jQuery.map( this, fn, until );

		if ( name.slice( -5 ) !== "Until" ) {
			selector = until;
		}

		if ( selector && typeof selector === "string" ) {
			matched …
base.js:94 Uncaught TypeError: this.parent.trigger is not a function
    at jQuery.fn.init.trigger (http://localhost:3000/app.bundle.js:26753:26)
    at jQuery.fn.init.jQuery.fn.(anonymous function) [as focus] (http://localhost:3000/app.bundle.js:12476:9)
    at Form.focus (http://localhost:3000/app.bundle.js:73475:31)
    at __WEBPACK_IMPORTED_MODULE_2__utils__.a.all.done.fail (http://localhost:3000/app.bundle.js:56617:14)
    at fire (http://localhost:3000/app.bundle.js:7503:31)
    at Object.add [as fail] (http://localhost:3000/app.bundle.js:7562:7)
    at Form.whenValidate (http://localhost:3000/app.bundle.js:56615:8)
    at Form.validate (http://localhost:3000/app.bundle.js:56590:32)
    at OrderForm._this.handleSubmit (http://localhost:3000/0.bdddcdd879b4404d0871.hot-update.js:138:25)
    at Object.ReactErrorUtils.invokeGuardedCallback (http://localhost:3000/app.bundle.js:32475:16)
```

br

simkea:
Okay, i believe the reason is deeper for this bug... pls delete this PR.
Is it possible to use parsley for a react component ?!...

marcandre:
Absolutely, all JS frameworks should be compatible with Parsley.

PR #1239, opened 2018-01-30
Fix accessibility : relationship between field and error message

stevenMouret:
I suggest to you an improvement for people with disabilities.  These changes allow to create relationship between the field and its error. When the user is on a field with an error, the assistive technology informs the user of the error.
Maybe _updateError function must be change also.

[You can found more infromation on the website of the W3C](https://www.w3.org/WAI/tutorials/forms/notifications/).

Best regards


marcandre:
Awesome, thanks!

PR #1241, opened 2018-01-31
Build dist after fix accessibility

stevenMouret:
Sorry guys I forgot to build dist folder.
Maybe it necessary to change the _updateError function. the idea is to add attribute aria-describedby on the field with value of the id of the error message and remove it when error message disapear.

I do not have much time now but I will try to improve accessibility.


marcandre:
No worries, I'll do a release soon. 
I also don't think it's necessary to change `_updateError`.

PR #1247, opened 2018-03-29
Remove duplicate license

veksen:
The homepage key is also duplicate, not sure if that's intentional, but I doubt the license was


marcandre:
Thanks!

Issue #1261, opened 2018-08-09
Compile parsley.js with webpack and babel issue


geosenna:
As there is some code into parlsey as such: 

> `' + this.name + '`

This lead to a syntax issue in IE11.

So i want to use Babel as I did for other modules I import, but when it come to parsley I get this issue:

```
ReferenceError: [BABEL] **/node_modules/parsleyjs/dist/parsley.js: Using removed Babel 5 option: **/node_modules/parsleyjs/.babelrc.blacklist - Put the specific transforms you want in the `plugins` option
```

And I can't translate parlsey code to Babel.

This is my webpack config:

```
module.exports = {
    entry: './app/js/app.js',
    mode: 'none',
    output: {
        filename: 'app.js',
        path: path.resolve(__dirname, 'public/scripts/')
    },
    module: {
        rules: [
            {
                test: /\.js$/,
                loader: 'babel-loader',
                query: {
                    presets: [['env', { modules: false }]]
                }
            }
        ]
    },
};
```

and my babelrc:

```
{
    "presets": [ "env" ]
}

```
Ay though on this?

marcandre:
Not too sure. It's possible we're using an old version of babel (and setup). Can you come up with PR to update `babel` and change the `babelrc.blacklist` correctly?

marcandre:
(check #1170 too)

therealshark:
@geosenna try to add ```babelrc: false```to the babel-loader options. This should fix the behaviour. The problem is, that the .babelrc from parsley gets read from the loader. By setting the option mentioned, that doesn't happen anymore :)

PR #1170, opened 2017-02-06
Update to babel 6

gempi:
- Update to babel 6

PR #1264, opened 2018-08-13
update Babel to the latest version

geosigno:
Babel has been updated to the latest version to avoid issue when compiling. ([#1261](https://github.com/guillaumepotier/Parsley.js/issues/1261))


marcandre:
Thanks for the PR :-)

I'm not sure why, but this has the same issue as #1170, namely that in case there is no `exports` or `define` (such as a browser), the code that loads it all changes from

```
(function (global, factory) {
  global.parsley = factory(global.jQuery);
})(this, function ($) {
```
to
```
(function (global, factory) {
  global.parsley = factory(global.jQuery);
})(undefined, function ($) {
```

So instead of doing `window.parsley = factory(window.jQuery)` (since `this` at the top level of browser is `window`), the new code does `undefined.parsley = factory(undefined.jQuery)` which won't work...

geosenna:
help needed

marcandre:
Ok, I'm looking into this.

marcandre:
I've spent hours so far trying to update this. It is really painful.

Progress so far: #1267

marcandre:
Merged in master

PR #1273, opened 2018-10-17
Improve greek translation for required fields.

torotil:
According to one of our clients the old error message means something like “the price is demanded” and doesn’t make a lot of sense. The new message can be translated as “required field”.

Perhaps this also applies similarly to the other error message translations, but I’ve only got a better translation for this one.


marcandre:
ευχαριστώ

PR #1280, opened 2019-01-08
Fix accessibility

stevenMouret:
- Fix aria-describedby when using data-parsley-error-message
- Manage aria-hidden attribute on error wrapper

### DIFF
--- a/src/parsley/ui.js
+++ b/src/parsley/ui.js
@@ -170,16 +170,28 @@ UI.Field = {
               .addClass('parsley-custom-error-message')
             );
 
-        return this._ui.$errorsWrapper
+        var replaceErrorMessage = this._ui.$errorsWrapper
           .addClass('filled')
+          .attr('aria-hidden', 'false')
           .find('.parsley-custom-error-message')
           .html(this.options.errorMessage);
+
+        var addDescribedby = this._ui.$errorClassHandler
+          .attr('aria-describedby', this._ui.errorsWrapperId);
+
+        return replaceErrorMessage && addDescribedby;
       }
 
-      return this._ui.$errorsWrapper
+      var removeErrorMessage = this._ui.$errorsWrapper
         .removeClass('filled')
+        .attr('aria-hidden', 'true')
         .find('.parsley-custom-error-message')
         .remove();
+
+      var removeDescribedby = this._ui.$errorClassHandler
+        .removeAttr('aria-describedby');
+
+      return removeErrorMessage && removeDescribedby;
     }
 
     // Show, hide, update failing constraints messages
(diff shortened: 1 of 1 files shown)


src/parsley/ui.js
```diff
@@ -170,16 +170,28 @@ UI.Field = {
...
+      var removeErrorMessage = this._ui.$errorsWrapper
         .removeClass('filled')
+        .attr('aria-hidden', 'true')
         .find('.parsley-custom-error-message')
         .remove();
+
+      var removeDescribedby = this._ui.$errorClassHandler
+        .removeAttr('aria-describedby');
+
+      return removeErrorMessage && removeDescribedby;
```
marcandre: This return statement makes no sense to me. Simplest change would be to first `removeAttr('aria-describedby')` then do as before, no?

marcandre:
Great, thanks for the PR. I commented on a `return`. Also, it would be great if this was part of one of the tests. Are you up to tweak this?

src/parsley/ui.js
stevenMouret: Bonjour Marc andré, 

I am not sure to understand. In the case there is no input assistance, it is necessary to remove aria-describedby. If we do not remove aria-describedby the technical assistance will try to read an empty ul.

marcandre: Please re-read carefully what I wrote. Nowhere do I propose not removing the `aria-describedby`.

stevenMouret: Hello Marc André,

Sorry this pull request is open for some time now.
Maybe it is possible to do more simple, but I am not a JS expert. The chunk of code you are highlighting just allow to remove the aria-describedby and the error message. If it is possible to make more simple do not hesitate to refactoring my code. If you need more explanation about the this accessibility changes I can explain more.

marcandre:
I've pushed a modified commit to master, but I'm running into issues on Travis.

I've wasted an hour of my life on incompatibilities that seem to me [completely ridiculous](https://stackoverflow.com/questions/46053414/npm-warn-requires-a-peer-of-but-none-is-installed-you-must-install-peer) but it's still not working.

stevenMouret:
> 
> 
> I've pushed a modified commit to master, but I'm running into issues on Travis.
> 
> I've wasted an hour of my life on incompatibilities that seem to me [completely ridiculous](https://stackoverflow.com/questions/46053414/npm-warn-requires-a-peer-of-but-none-is-installed-you-must-install-peer) but it's still not working.

I think you're getting the wrong issue. My changes do not affect dependencies.

marcandre:
> My changes do not affect dependencies.

Indeed, there is no doubt about that.

> I think you're getting the wrong issue. 

No. I'm just explaining where I'm at. I have never thought or implied that your commit, or my modified version of it, was responsible for the troubles on Travis.

marcandre:
Merged in with modifications as a1a5d36

Should be fixed in 2.9.2

Thanks again for the PR :+1:

PR #1281, opened 2019-01-25
add error messages color

SakhriHoussem:


marcandre:
Sounds like a good idea. Thanks

PR #1282, opened 2019-01-25
Replace ‘Gender’ radio buttons in simple example.

matthewmcvickar:
This commit replaces the 'Gender' radio button field on the 'Simple form example' page. There are a few reasons for this:

- Gender is not binary, so offering only choices for 'male' and 'female' doesn’t reflect users’ experience and reality.
- [Asking for users' gender is problematic and unnecessary](https://uxdesign.cc/designing-forms-for-gender-diversity-and-inclusion-d8194cf1f51) in most cases, so let’s discourage its use as a common field.


marcandre:
Thanks

PR #1295, opened 2019-03-26
euvatin validator and some other minor changes

MarcoSulla:


src/i18n/en.extra.js
```diff
@@ -6,9 +6,9 @@ Parsley.addMessages('en', {
   minwords: "This value is too short. It should have %s words or more.",
   maxwords: "This value is too long. It should have %s words or fewer.",
   words:    "This value length is invalid. It should be between %s and %s words long.",
-  gt:       "This value should be greater.",
-  gte:      "This value should be greater or equal.",
-  lt:       "This value should be less.",
-  lte:      "This value should be less or equal.",
-  notequalto: "This value should be different."
+  gt:       "This value should be greater of %s.",
```
marcandre: greater *than*, less *than*, equal *to*

src/parsley/validator_registry.js
```diff
@@ -359,7 +359,16 @@ ValidatorRegistry.prototype = {
           return value === refOrValue;
       },
       priority: 256
-    }
+    },
+    euvatin: {
```
marcandre: Check how `integer` is implemented. You don't need an actual function.

marcandre:
Added a few comments. Also you have a failing test.

src/parsley/validator_registry.js
MarcoSulla: I tried but it fails integration tests with an obscure error.

MarcoSulla:
Done.

test/unit/validator_registry.js
```diff
@@ -191,6 +191,15 @@ describe('ValidatorRegistry', () => {
     expect($('#element').psly().isValid()).to.be(true);
     $('#equalto').remove();
   });
+  it('should have a euvatin validator', () => {
+    expectValidation('',    'euvatin').to.be(true);
```
marcandre: This test is wrong. I'd simply remove it.

MarcoSulla:
In teory it should not pass only if the field is also required?

marcandre:
Validating `""` is not the concern of the regex. Best not test for it.

MarcoSulla:
I removed it but it continues to have the obscure message.

PS: it's a bit confusing that some validators needs the empty string check for `required` compatibility and others not, anyway.

marcandre:
I apologize. Your implementation was correct and I was confused asking you to check `integer` validation. I simply went back to your original implementation and merged it to master.
Great work, thanks again

PR #1296, opened 2019-03-27
Create date.js

MarcoSulla:


marcandre:
I just remembered now, but there's already `type == "date"`. How does this differ?

MarcoSulla:
It's useful for libraries like bootstrap datetimepicker that uses type=text.
Furthermore you can pass a specific format. type=date has only a simpler pattern attribute, and the extra dateiso (that is not documented) does not have a format parameter.

PS: I added this code at the last moment to datebeforenow and dateafternow:

        var valid_date = Parsley._validatorRegistry.validators.date.validateString(value, format);

        if (! valid_date) {
            return true;
        }

My intention is to trigger the date validator. If the date is invalid, the invalid date message of data-parsley-date should be displayed.
Is this correct? I'm not sure it's a good idea, since I'm using a private attribute, and furthermore I'm not sure the message will be displayed.

marcandre:
> It's useful for libraries like bootstrap datetimepicker that uses type=text.

You can still use `type=text data-parsley-type=date`

> Furthermore you can pass a specific format.

Not sure what you mean

MarcoSulla:
> > Furthermore you can pass a specific form
> Not sure what you mean

`date` and the other validators are intended to be used in this way, for example:
`data-parsley-date="YYYY-MM-DD"`

The format is the `moment` format. The date is passed to the `moment` constructor with the format as parameter and the strict parameter set to `true`. If the date is not in that specific format, the error is displayed. See [the moment docs about it](https://momentjs.com/docs/#/parsing/string-format/).

What about the calling of `date` validator inside the other validators?

MarcoSulla:
It seems the validator `date` called inside `datebeforenow` and `dateafternow` does not show any error.

Can you please tell me how can I invoke a validator from another validator?

MarcoSulla:
I opened a [question on Stackoverflow](https://stackoverflow.com/questions/55521887/is-it-there-a-way-to-call-a-parsley-validator-inside-another-one/55540692#55540692), can you please help me?

MarcoSulla:
Please, can you tell me if you want to add it to the project?

marcandre:
It's fine to add to `/extra`, but simply repeat the call to `isValid` from `datebeforenow` and `dateafternow`...

MarcoSulla:
So there's no possibility to call `date` inside `datebeforenow` and the other validator, and show the error message of `date` if the date is wrong?

MarcoSulla:
Well, I've done `date.isValid()` check as you suggested. I hope that now they can be included in `extra`.

src/extra/validator/date.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+window.Parsley.addValidator('datebeforenow', {
+    validateString: function(value, format) {
+        if (! value) {
+            return true;
+        }
+        
+        if (! date.isValid()) {
+            return true;
```
marcandre: Shouldn't that be `return false`??

MarcoSulla: No, it must return `true` because you have also to add `data-parsley-date`, or you'll get an erroneous message. That's because I was trying to launch `data-parsley-date` inside the other two validators.

MarcoSulla: Think it as `required` field. If another validator gets an empty value, it returns `true`.

marcandre: Best return `true`. `data-parsley-date` should simply have a higher priority. If you have both `data-parsley-date` and `datebeforenow`, you will get different error messages for an invalid date and a valid date that is after now. If you only have `datebeforenow`, then you will get a single error message and you will get it for an invalid date or a date after now.

MarcoSulla: > If you have both `data-parsley-date` and `datebeforenow`, you will get different error messages for an invalid date and a valid date that is after now

Well, this is exactly what I want. I think it's more precise. But if you prefer a single parser and message, I'll modify them.

Anyway I think it will be more, more powerful if you could call another validator inside the validator. You should think about adding this possibility.

MarcoSulla: Done.

marcandre: Cool. Do you know how to squash your commits?

MarcoSulla: Mh, no. How can I do?

marcandre: If you want to learn, google is your friend. For example: https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git/5201642#5201642

MarcoSulla: Mh, I preferred to undo all and redo it. For three files...

See #1304

PR #1304, opened 2019-05-04
Added data-parsley-date and friends

MarcoSulla:


marcandre:
Merged! Thanks for the PR

Note: I squashed the commits myself, since they weren't independent. BTW, you had trailing whitespaces, you should fix the settings of your code editor to avoid this.

MarcoSulla:
Thank you! I prefer to leave trailing whitespaces, so if I want to add some code in that line, it's already indented :)

PR #1320, opened 2019-11-06
Create datecomparison.js

SakhriHoussem:
# Create date comparison validator
I find a problem when I use **parsley.js** to compare between two inputs with type **month** or **date**
```html
<input type="month" id="from" name="from">
```
or
```html
<input type="date" id="from" name="from">
```
the problem is that **parsley.js** compares just between years but it doesn't take into consideration days or months when it compares 
## for example :
I want to make sure that the date in the input **from** must be near than the input **to**
```html
<input type="month" id="from" name="from">
<input type="month" id="to" name="to">
```
so I write these two functions to try to fix this problem

### DIFF
--- a/src/extra/validator/datecomparison.js
+++ b/src/extra/validator/datecomparison.js
@@ -1,0 +1,54 @@
+// Load this after Parsley for additional date comparison validators
+// Note: comparing with a reference isn't well supported and not recommended.
+// import jQuery from 'jquery'; // Remove this line in ES3
+
+// date-gt, date-lt extra validators
+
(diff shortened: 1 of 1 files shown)


marcandre:
Hi! Thanks for the PR.

For the first time if my life, I don't have my computer with me, as I'm trekking the Annapurna circuit! I will be reunited with my laptop in about two weeks; I'll either look at this on a plane or when I'm back home in early December...

I don't recall anything about `type=month`, so I'm also wondering if Parsley couldn't deal with this better, without the need of an extra validator...

Issue #1100, opened 2016-04-27
Vanilla JS version


zslabs:
What kind of issue is this? (put 'x' between the square brackets)

 - [x ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use
       http://stackoverflow.com/questions/ask .
       Provide working code, starting from http://codepen.io/marcandre/pen/jqbzyN?editors=101.
       We monitor the tag `parsley.js`.

 - [ ] Bug report. If you’ve found a bug, you must provide a minimal example in a CodePen,
       starting from http://codepen.io/marcandre/pen/jqbzyN?editors=101 .

 - [x] Feature Request. Make sure there's no good way to do what you want first;
       consider asking on http://stackoverflow.com/questions/ask first.

Hi,
Are there any specific features tying this to jQuery nowadays? Is there any interest in converting this to a vanilla JS version?

marcandre:
To be very honest, I don't know for absolutely certain, as I never write JS without jQuery, so my knowledge of vanilla DOM is very limited. 

I personally have zero interest in a vanilla `parsley`. Let me try to me more explicit about why.

I don't personally write without jQuery and I don't plan to. Nor do I write in HTML, ES3/5, or in CSS for that matter. I consider HTML, JS, Dom & CSS like assembly language, a platform to target but requiring an essential layer that provides a decent interface to them (slim/Haml, CoffeeScript/ES6, jQuery & Sass/less respectively). Of course that's a personal opinion, but hey, as the only active maintainer, it counts :smile:

More practically, from my limited knowledge:
I don't know at which version of IE one could start adding attributes to native objects (ideally without memory leaks). We use jQuery `data` so that's a solved issue for us, so we can support IE8+.

`Parsley` makes extensive uses of promises, although we could use ES6's instead. It uses `$.ajax`, which has a *much* nicer interface in `jQuery` than the vanilla equivalent, but again that could be kind of done "manually". 

More importantly, there are two things that can't be done reasonably without `jQuery` to provide easy options:

#### Config of ajax

It wouldn't be reasonable to provide compatibility with ajax options and ajax validators for the `remote` validator. These are `jQuery` specific and are not easy to duplicate.

#### Selectors.

[...]

PR #1322, opened 2019-12-05
Avoid jQuery dependency in some places [#1100]

bububo:


src/parsley/main.js
```diff
@@ -113,11 +113,10 @@ $.each('addError updateError'.split(' '), function (i, method) {
 // ### PARSLEY auto-binding
 // Prevent it by setting `ParsleyConfig.autoBind` to `false`
 if (false !== window.ParsleyConfig.autoBind) {
-  $(function () {
```
marcandre: This changes the timing of the autobind. It should be remain the same, i.e. after loading is complete instead of instantly.

bububo: Done. Please review

marcandre:
This fell through the cracks, sorry.

Merged, at last.

PR #1326, opened 2020-01-15
[TASK] Add macedonian and icelandic language

joey-beech-it:
We needed the Macedonian and Icelandic language in a project, got it translated.


marcandre:
Thanks! :+1:

PR #1332, opened 2020-05-05
Update gulp patterns for v4

phated:
Hey hey @marcandre! I saw this gulpfile when getting you set up for Tidelift and figured I'd jump in to help align it with the current recommended patterns for gulp 4.

There's still a bunch of improvements that could be made, like streamlining the function names to avoid all the `displayName` annotations and cleaning up the docco task, but I think this is a great start.


marcandre:
Wow, looks great! Thanks ❤️

PR #1338, opened 2020-08-26
master

jongfighter:
#Why
Fix incorrect translation in number and digits

#What and How
First, previous translation says number is phone number. 
It seems that the first committer used Google translator's first recommendation.
However, in here, the third recommendation is correct, so I fix it.
Second, digits means characters between 0 and 9. 
Because translatied word for number and digit is identical in korean everday language,
ithe description is ambiguous.
Though there is math vocabulary for digits and number, it is not common to public.
Thus, I added an explanatory description to remove ambiguity.

### DIFF
--- a/src/i18n/ko.js
+++ b/src/i18n/ko.js
@@ -4,11 +4,11 @@ import Parsley from '../parsley';
 Parsley.addMessages('ko', {
   defaultMessage: "입력하신 내용이 올바르지 않습니다.",
   type: {
-    email:        "입력하신 이메일이 유효하지 않습니다.",
-    url:          "입력하신 URL이 유효하지 않습니다.",
-    number:       "입력하신 전화번호가 올바르지 않습니다.",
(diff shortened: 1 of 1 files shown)


jongfighter:
> #Why
> Fix incorrect translation in number and digits
> 
> #What and How
> First, previous translation says number is phone number.
> It seems that the first committer used Google translator's first recommendation.
> However, in here, the third recommendation is correct, so I fix it.
> Second, digits means characters between 0 and 9.
> Because translatied word for number and digit is identical in korean everday language,
> ithe description is ambiguous.
> Though there is math vocabulary for digits and number, it is not common to public.
> Thus, I added an explanatory description to remove ambiguity.

Last, the other ones are thought-for-thought translation, but it is better to translate word-for-word translation.
Thus, I fix it.

marcandre:
Thanks!

PR #1339, opened 2020-09-05
docs: Fix simple typo, querry -> query

timgates42:
There is a small typo in dist/parsley.js, doc/assets/spec-build.js, src/parsley/remote.js.

Should read `query` rather than `querry`.


marcandre:
Thankks 👍

PR #1342, opened 2020-10-21
fix: Locale - Icelandic translations

OArnarsson:
Stumbled upon the badly translated error messages and replaced them.


marcandre:
Thanks!

PR #1345, opened 2020-11-28
Updating Localization for Indonesian language

Mammo32:
I have make change of translating in src/i18n/id.js and dist/i18n/id.js


Mammo32:
I have updating localization files

Mammo32:
I have updating localization files

PR #1379, opened 2026-03-18
Fix Romanian validation messages for clarity

vladutilie:
An extra placeholder broke the message, now it should be fine.


marcandre:
Thank you

